PR #2, opened 2022-12-19
Factor out providers

mogul:
[Reusable modules should not configure their own providers, only state their requirements.](https://developer.hashicorp.com/terraform/language/modules/develop/providers#providers-within-modules) Providers should be configured (and if necessary, explicitly bound) by the calling module.

Modules should also not use `~>` when specifying their provider requirements, because new major versions may not break backward compatibility.

### DIFF
--- a/clamav/variables.tf
+++ b/clamav/variables.tf
@@ -1,20 +1,3 @@
-variable "cf_api_url" {
-  type        = string
-  description = "cloud.gov api url"
-  default     = "https://api.fr.cloud.gov"
-}
-
-variable "cf_user" {
-  type        = string
-  description = "cloud.gov deployer account user"
-}
-
-variable "cf_password" {
-  type        = string
-  description = "secret; cloud.gov deployer account password"
-  sensitive   = true
-}
-
 variable "cf_org_name" {
   type        = string
   description = "cloud.gov organization name"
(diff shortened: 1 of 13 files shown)


clamav/variables.tf
```diff
@@ -25,14 +8,9 @@ variable "cf_space_name" {
   description = "cloud.gov space name (staging or prod)"
 }
 
-variable "app_name" {
-  type        = string
-  description = "base application name to be scanned by clamav"
-}
-
-variable "env" {
+variable "name" {
```
rahearn, line 11: [suggestion] this change needs a matching one within `clamav/main.tf` to update both the app's name and the route being used.

It'll be a breaking change, but I think that's ok? I'm not aware of any projects that are both using this module and also currently deploying clamav.

rahearn (changes requested):
Thanks, this is a lot cleaner. Just the one change for the clamav variables needed.

clamav/variables.tf
rahearn: Or, even better, back out these changes from this PR and include them in #3

mogul: I'll do that, must have missed this when I split the branch the first time. ;)

rahearn (approved):
🎉

PR #3, opened 2022-12-19
Remove assumptions about how the service will be used

mogul:
**NOTE: THIS IS A BREAKING CHANGE**

Modules shouldn't know anything about environments or what applications the services might be bound to. The calling module should specify the name using its own context/convention, rather than the module making assumptions.


mogul:
(Moved the breaking change from #2 over here.)

rahearn (approved):

PR #9, opened 2023-03-01
Bump suggested clamav-rest version

mogul:
See https://github.com/ajilach/clamav-rest/issues/15. Technically this version is not required for module version 0.3.0, but a lot of people are going to just blindly copy the suggested tag here so we may as well start them from that one.


rahearn (approved):
Good idea

PR #14, opened 2023-05-04
Accept an https_proxy to use for ClamAV definition updates

mogul:


rahearn (approved):
🎉

mogul:
This isn't yet working because

1. If we have the env var set while the Docker image is staging, then CF can't reach Docker Hub and errors out.
2. 1. `freshclam` doesn't respect the env var natively... We have to set it in the `freshclam.conf` file.

So we're going to [make a PR upstream that addresses these problems](https://github.com/GSA-TTS/FAC/issues/1064#issuecomment-1536820002).

PR #16, opened 2023-05-04
Update README to reference most recent tag

mogul:
I'm not sure if there's a better way to do this than to manually make a PR every time... Any ideas?


rahearn:
We could change each of the examples to `ref=main` (to keep it working no matter what) or `ref=TAG_NAME` (to make it obvious that it needs to be changed) and then include a blurb somewhere to point people to the releases page to get the actual version to use?

rahearn:
This appears to be rendered moot by #17

PR #19, opened 2023-05-11
Use rolling strategy for deployment

mogul:
Now that the cloudfoundry_app resource uses CAPI v3 and the "rolling" strategy is supported, we can just use it and not worry about the "blue-green" issues that the id_bg attribute was intended to address.


rahearn (approved):
FYI, I recently moved the us-notify SSBs from blue-green to rolling, and they both failed on the first deploy via terraform. No real idea why, but they seem to be fine after running the terraform again.

Issue #15, opened 2023-05-04
Create module tests


rahearn:
Incorporate [new testing features](https://developer.hashicorp.com/terraform/language/modules/testing-experiment) for CI/CD tests for these modules.

[Similar examples in use on the FAC project](https://github.com/GSA-TTS/FAC/tree/main/terraform/shared/modules/egress-proxy/test)

rahearn:
New link for terraform test docs: https://developer.hashicorp.com/terraform/language/tests

PR #23, opened 2023-05-18
Add test for S3, database, and redis modules

rahearn:
#15 

**Potential issue:** need to be able to actually create the resources, so might not be able to run the bigger / multi-space modules within sandbox-gsa

`cg_space` (permission issues), `clamav` (memory quota too small), and `domain` (relies on manual steps and DNS entries) are all unable to be tested in this manner within `sandbox-gsa`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,9 +12,9 @@ jobs:
     name: Integration test
     strategy:
       fail-fast: false
-      max-parallel: 1
+      max-parallel: 2
       matrix:
-        module: ["s3"]
+        module: ["s3", "database", "redis"]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,7 @@ jobs:
     name: Integration test
     strategy:
       fail-fast: false
+      max-parallel: 1
       matrix:
         module: ["s3"]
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: Integration test
+    strategy:
+      fail-fast: false
+      max-parallel: 1
```
rahearn, line 15: since these are integration tests, I wanted to avoid hammering cloud.gov with a bunch of parallel runs. We may want to increase this a bit though, especially as the database and redis modules come under test, since they take a while to create

rahearn:
notes from conversation with @mogul:

1. look into using external data source to input the org/space needed for tests
2. look into using external data source, or other data sources, to run more involved test assertions

mogul:
> notes from conversation with @mogul:
> 
> 1. look into using external data source to input the org/space needed for tests
> 2. look into using external data source, or other data sources, to run more involved test assertions

I pushed [the branch where I'm proving out these methods](https://github.com/GSA-TTS/FAC/tree/use-terraform-test). They work!

rahearn:
`terraform test` has undergone major changes since this PR was originally opened. I'm closing this in favor of a green-field approach to testing the modules.

PR #24, opened 2023-05-31
Update environment to support proxy

asteel-gsa:
After creating a fork of clamav inside the GSA-TTS Org, we are proposing the following changes so that we may pass the environment variables necessary for freshclam proxy.

In parallel, we would be refactoring our https-proxy to now [output these items](https://github.com/GSA-TTS/FAC/pull/1174/files) specifically as:

```tf
  https_proxy    = module.https-proxy.https_proxy
  proxy_port     = module.https_proxy.port
  proxy_username = module.https_proxy.username
  proxy_password = module.https_proxy.password
```

to be consumed here, and set the ENV vars for the following items [updated entrypoint.sh](https://github.com/GSA-TTS/clamav-rest/blob/9a388c8d4aa8815797ba1c0b9521ae1fe7f6ed55/entrypoint.sh#L19-L32):
```sh
$PROXY_SERVER
$PROXY_PORT
$PROXY_USERNAME
$PROXY_PASSWORD
```

### DIFF
--- a/clamav/main.tf
+++ b/clamav/main.tf
@@ -33,8 +33,11 @@ resource "cloudfoundry_app" "clamav_api" {
     # Only set "https_proxy" if a value was supplied.
     # Otherwise, ensure that a harmless envvar gets set instead.
     # This avoids confusing the app with an https_proxy that's set to ""!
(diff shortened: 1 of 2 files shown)


mogul:
Closing in favor of #25, which makes some edits that I couldn't figure out how to do based on your fork.

PR #27, opened 2023-06-12
Support hostname when setting up a route+domain service

mogul:
For example, one might want their app mapped to the FQDN "app.fac.gov", not just "fac.gov". 😉


mogul:
I'm about 90% sure that this will work... It's hard to test easily without passing something all the way through our pipeline.

akf (approved):
Looks reasonable.

pauldoomgov (approved):
This looks good to me, for what it is worth!

PR #28, opened 2023-09-14
Enable using these modules with version constraints

mogul:
Adds a `semver` module that will calculate a tag version from this repository using NPM-style version constraints.

This does not update the main README's examples to use this new semver method; we should probably discuss how best to do that before merging, hence the draft state of the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # terraform-cloudgov
 
-Terraform modules for cloud.gov managed services commonly used by [18f/rails-template](https://github.com/18f/rails-template) based apps
+Terraform modules for working with cloud.gov commonly used by [18f/rails-template](https://github.com/18f/rails-template) based apps
 
 ## Module Examples
 
@@ -117,3 +117,7 @@ module "egress_space" {
   ]
 }
 ```
+
+## Staying up to date
+
+Note that [Terraform doesn't support version constraints for github module sources](https://developer.hashicorp.com/terraform/language/modules/sources#github). To use the latest available versions of these modules using a version constraint, use our [semver module](./semver). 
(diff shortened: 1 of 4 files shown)


rahearn (commented):
I like this idea, and I think for the main README we could probably do what you're doing in `semver/README.md` with the locals block either saying "FILL IN CONSTRAINTS" or doing something like `~>0.7` that will get all updates up to 1.0 and we can keep that more or less constant until we have breaking changes.

mogul:
So each stanza in the README would look something like this? (Note I'm _terrible_ at understanding version constraints, so I'm kind of guessing based on the sparse examples [here](https://semver.npmjs.com/#syntax-examples). And note this module uses NPM's version constraints, not Terraform's!)

```
module "latest-non-breaking-version" {
  source             = "github.com/18f/terraform-cloudgov//semver"
  version_constraint = "~0.7.0"
}

module "database" {
  source = "github.com/18f/terraform-cloudgov//database?ref=v${module.latest-non-breaking-version.target_version}"
  # [...]
}
```

(I also think it might be a good time to start moving those stanzas into a README.md in each subdirectory.)

mogul:
Oh oh, wait, you meant actually using the exact same pattern with the `for_each` at the top of the main README.md! OK, let me add a commit for that.

rahearn:
> Oh oh, wait, you meant actually using the exact same pattern with the `for_each` at the top of the main README.md! OK, let me add a commit for that.

I did mean that, but also I'm good with either way

rahearn (approved):
Love that this is a purely optional quality of life add-on

mogul:
Welp, uh, [bad news](https://registry.terraform.io/modules/rhythmictech/find-release-by-semver)... You can use this module's output as the `source` value for another module. Doh! Expect a PR rolling out the documentation changes.

PR #29, opened 2023-12-06
Add tags to modules

asteel-gsa:
This PR adds tags as an optional variable, defaulting to an empty list. 

Working on another task, the notion of using a tag to retrieve an item from `$VCAP_SERVICES` arose and would like to propose this change.

```terraform
module "module-name" {
  source = "github.com/18f/terraform-cloudgov//{module-source}"
  [...]
  tags             = ["tag1", "tag2", "..."]
} 
```

### DIFF
--- a/clamav/main.tf
+++ b/clamav/main.tf
@@ -26,7 +26,6 @@ resource "cloudfoundry_app" "clamav_api" {
   timeout      = 600
   strategy     = "rolling"
   docker_image = var.clamav_image
-  tags         = var.tags
   routes {
     route = cloudfoundry_route.clamav_route.id
   }
--- a/clamav/main.tf
+++ b/clamav/main.tf
@@ -26,6 +26,7 @@ resource "cloudfoundry_app" "clamav_api" {
   timeout      = 600
   strategy     = "rolling"
   docker_image = var.clamav_image
+  tags         = var.tags
   routes {
     route = cloudfoundry_route.clamav_route.id
   }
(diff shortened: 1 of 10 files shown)


asteel-gsa:
Can't seem to add reviewers. CC: @rahearn @mogul

rahearn (changes requested):
I like it. My only request is that you add it to the other modules as well.

asteel-gsa:
> I like it. My only request is that you add it to the other modules as well.

Sure thing! Give me just a moment to add it!

clamav/main.tf
```diff
@@ -26,6 +26,7 @@ resource "cloudfoundry_app" "clamav_api" {
   timeout      = 600
   strategy     = "rolling"
   docker_image = var.clamav_image
+  tags         = var.tags
```
rahearn, line 29: It looks like [tags does not exist on the cloudfoundry_app resource](https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs/resources/app) so please remove from `clamav`

rahearn (changes requested):
thanks for adding to the other service instance resources, and sorry for the overly broad change request on the last review.

clamav/main.tf
asteel-gsa: Fixed! Good catch

rahearn (approved):
🎉

rahearn:
Released as `v0.7.1`

PR #32, opened 2024-03-01
Remove semver module

mogul:
Reference [this comment](https://github.com/18F/terraform-cloudgov/pull/28#issuecomment-1973892187)... This was a good idea that will not actually work because Terraform wants a static `source` value for modules.

I've updated the README.md to refer to `v1.0.0`, which is what we should tag once this is merged... Technically I broke backward compatibility when I removed the module, and I think it's time we start using a major version number to signal when that happens.

---

Reverts "Merge pull request #28 from 18F/add-semver-module"
This reverts commit 0b02d6072e1c6cb115bfc5d2b60d8ce4f5662237, reversing changes made to 7db01d5283148de2bdf2038774ebdea3ed049268.


mogul:
@rahearn would you mind approving and merging this?

rahearn:
> @rahearn would you mind approving and merging this?

sorry - just getting back into this repo. Before I approve, what is the intended use of the `find-release-by-semver` if not for what you tried to do?

mogul:
> what is the intended use of the `find-release-by-semver` if not for what you tried to do?

As an example, one might write HCL that uses that module to divine the newest gitref matching the supplied semver, then [use that gitref to supply a .zip of the downloaded release to a `cloudfoundry_app` resource](https://github.com/GSA-TTS/FAC/blob/356f84552b39b3631a0653577f9098c6b47af9c0/terraform/shared/modules/https-proxy/https-proxy.tf#L46-L60).

rahearn (approved):
Approved. I'm going to hold off on tagging anything quite yet and do a review of anything else we might want to get into that "1.0.0" release, since that's a bigger release than just a new major version.

rahearn:
1.0.0 milestone tracker: https://github.com/18F/terraform-cloudgov/milestone/1

PR #33, opened 2024-03-18
Adds `json_params` as an input

asteel-gsa:
Instead of manually doing `cf update-service ${service_name} -c '{"key":"value"}'` lets do it with terraform allowing us to specify additional arguments in json format.
```terraform
> jsonencode({"storage" : 10,})
"{\"storage\":10}"
```

### DIFF
--- a/s3/variables.tf
+++ b/s3/variables.tf
@@ -35,7 +35,7 @@ variable "tags" {
 variable "json_params" {
   description = "A JSON string of arbitrary parameters"
   type        = map(string)
-  default     = false
+  default     = null
   # See options at https://cloud.gov/docs/services/s3/#setting-optional-parameters
 }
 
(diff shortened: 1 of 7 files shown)


asteel-gsa:
@rahearn @mogul for your review.

mogul (approved):

s3/variables.tf
```diff
@@ -31,3 +31,11 @@ variable "tags" {
   type        = list(string)
   default     = []
 }
+
+variable "json_params" {
+  description = "A JSON string of arbitrary parameters"
+  type        = map(string)
```
rahearn, line 37: [question] is there a reason this one is `map(string)` and the other two are `map(object)`?

mogul: Getting fixed in [this other PR](https://github.com/18F/terraform-cloudgov/pull/34)...

PR #34, opened 2024-03-18
Make JSON conversion the user's responsibility

asteel-gsa:
https://stackoverflow.com/a/64433674

```
  │ The object type constructor requires one argument specifying the attribute
  │ types and values as a map.
```

### DIFF
--- a/database/variables.tf
+++ b/database/variables.tf
@@ -33,7 +33,14 @@ variable "tags" {
 
 variable "json_params" {
   description = "A JSON string of arbitrary parameters"
-  type        = map(object)
+  type        = map(object({
+    storage = number
+    backup_retention_period = number
+    binary_log_format = string
+    enable_pg_cron = bool
+    enable_functions = bool
+    storage_type = string
+  }))
   default     = null
   # See options at https://cloud.gov/docs/services/relational-database/#setting-optional-parameters-1
 }
--- a/database/variables.tf
+++ b/database/variables.tf
@@ -33,14 +33,7 @@ variable "tags" {
 
 variable "json_params" {
   description = "A JSON string of arbitrary parameters"
-  type = map(object({
-    storage                 = number
-    backup_retention_period = number
-    binary_log_format       = string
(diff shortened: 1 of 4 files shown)


database/variables.tf
```diff
@@ -33,7 +33,14 @@ variable "tags" {
 
 variable "json_params" {
   description = "A JSON string of arbitrary parameters"
```
mogul, line 35: I missed that the actual parameter is expected to be the map, rather than the string! In that case we should definitely update the description here.

redis/variables.tf
```diff
@@ -32,7 +32,7 @@ variable "tags" {
 
 variable "json_params" {
   description = "A JSON string of arbitrary parameters"
```
mogul, line 34: Same feedback on description here.

mogul (changes requested):
Update the descriptions

database/variables.tf
```diff
@@ -33,7 +33,14 @@ variable "tags" {
...
-  type        = map(object)
-  default     = null
+  type = map(object({
```
mogul, line 36: Will this require specifying all of these elements? Or will they remain optional?

mogul:
The README.md should be adjusted too... It shows passing a JSON string to the variable. It should instead strip out the call to `jsonencode` since the variable expects a map. And the module should do the `jsonencode` internally, which it's not doing now!

mogul:
More thoughts on that: If the module is expected to map its Terraform-native map variable to the json, then the modules' variable names should just be `params`, not `json_params`.

mogul:
You should change the other modules to specify the shape of the map for consistency:
https://github.com/asteel-gsa/terraform-cloudgov/blob/c38af82f07190734589a83edf932c8220c23b2f9/s3/variables.tf#L37
https://github.com/asteel-gsa/terraform-cloudgov/blob/c38af82f07190734589a83edf932c8220c23b2f9/redis/variables.tf#L35

mogul (approved):

PR #39, opened 2024-05-08
Optionally allow specifying 0:n target apps for domain service

### LINKED ISSUE #12: domain module should expect 0:n apps
The interface of the `domain` module is a little confusing because the module _requires_ an app as variable input, and that input is expected to be singular. 

- The module only takes one app as an argument, but [a `cloudfoundry_route` can target multiple apps](https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs/resources/route#target). So really that variable shape should be a _set_ of app IDs, not just one, and the resource should have [a `dynamic` block](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks) for iterating over them to specify `target` sub-blocks.
- An app isn't really required in order to create a route in CF. A domain and route could be created in isolation, without being bound to any app, and then bound to an app that gets created later/elsewhere (eg outside of Terraform) via `cf map-route`. In Terraform, mapping a route can be done by specifying [the `routes` attribute on the `cloudfoundry_app` resource](https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs/resources/app#routes) when it's created. So there should be a default of no apps to bind here. Given the point above, that would mean the variable should have a default value that's an empty set. The `dynamic` block would then create no `target` sub-blocks, which is what we want.

I could address these, but I'm not currently using the module myself and I don't want to break anything that others are using. So I'm noting the issue and resolution here in case someone else wants to take care of it!

rahearn:
Closes #12 

Created this as a draft to get feedback on while also testing the upgrade path with a real app.

### DIFF
--- a/domain/variables.tf
+++ b/domain/variables.tf
@@ -10,7 +10,14 @@ variable "cf_space_name" {
 
 variable "app_name_or_id" {
   type        = string
-  description = "base application name to be accessed at this domain name"
+  description = "base application name or id to be accessed at this domain name. Conflicts with var.app_names_or_ids"
+  default     = null
+}
+
+variable "app_names_or_ids" {
+  type        = list(string)
+  description = "base application names or ids to be accessed at this domain name. Overwritten by var.app_name_or_id"
+  default     = []
 }
 
 variable "name" {
--- a/domain/variables.tf
+++ b/domain/variables.tf
@@ -10,13 +10,13 @@ variable "cf_space_name" {
 
 variable "app_name_or_id" {
   type        = string
-  description = "base application name or id to be accessed at this domain name. Conflicts with var.app_names_or_ids"
+  description = "base application name or id to be accessed at this domain name. Overrides var.app_names_or_ids if used"
   default     = null
 }
 
 variable "app_names_or_ids" {
   type        = list(string)
-  description = "base application names or ids to be accessed at this domain name. Overwritten by var.app_name_or_id"
+  description = "base application names or ids to be accessed at this domain name. Overridden by var.app_name_or_id if used"
   default     = []
 }
 
--- a/domain/variables.tf
+++ b/domain/variables.tf
@@ -22,7 +22,7 @@ variable "app_names_or_ids" {
 
 variable "name" {
   type        = string
-  description = "name of the service instance"
+  description = "name of the service instance. Required if not passing in app names or ids"
   default     = ""
 }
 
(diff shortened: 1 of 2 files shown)


domain/variables.tf
```diff
@@ -10,12 +10,19 @@ variable "cf_space_name" {
...
 variable "app_name_or_id" {
   type        = string
-  description = "base application name to be accessed at this domain name"
+  description = "base application name or id to be accessed at this domain name. Conflicts with var.app_names_or_ids"
+  default     = null
+}
+
+variable "app_names_or_ids" {
+  type        = list(string)
+  description = "base application names or ids to be accessed at this domain name. Overwritten by var.app_name_or_id"
```
mogul, line 19:
```suggestion
  description = "base application names or ids to be accessed at this domain name. Overridden by var.app_name_or_id if used"
```

domain/variables.tf
```diff
@@ -10,12 +10,19 @@ variable "cf_space_name" {
 
 variable "app_name_or_id" {
   type        = string
-  description = "base application name to be accessed at this domain name"
+  description = "base application name or id to be accessed at this domain name. Conflicts with var.app_names_or_ids"
```
mogul, line 13:
```suggestion
  description = "base application name or id to be accessed at this domain name. Overrides var.app_names_or_ids if used"
```

mogul (approved):
I'm approving, but I don't think the trade-off between the single/plural variable options is clear from the documentation strings. As it is, they say that the singular version has primacy, but I think that should be reversed... I think if you say that the singular one is deprecated and the plural one is new and supported going forward, then it's more obvious that the deprecated single method is a subset of what you can do with the new plural method. (It doesn't change the backward compatibility test that you have at main.tf:4; folks not ready to use the new var will still have a non-null value for the singular variable so nothing will break for them... though we should see if there's a way to warn them about the deprecation if they're using it).

rahearn:
> though we should see if there's a way to warn them about the deprecation if they're using it

I'm not yet convinced that this should be a deprecation. I think specifying 0 or 1 app is the obvious benefit, and specifying more than 1 is a very niche behavior that I'm not convinced anyone is actually using. The docs say it's for load balancing, but 99% of cases load balancing would be done via horizontal scaling of a single app, not by adding additional apps to your route.

Given that, I don't love forcing everyone to remember that they have to pass that option as an array moving forward.

mogul:
Right, I was going to suggest that the default behavior and documented example would just be to pass an array, even if it just has the one app in it. If you don't agree, then it's just a matter of taste. I'm not overseeing any projects using this module, so I certainly defer to you!

rahearn:
> default behavior and documented example would just be to pass an array

I think that's what I'd do if I was writing the module from scratch today, but since it's niche behavior and we have existing users I think I'll stick with supporting the singular version as preferred. We can always revisit that decision in the future and deprecate/remove at that point.

anna-m-gsa (approved):

PR #48, opened 2024-06-18
Add module tests

### LINKED ISSUE #15: Create module tests

rahearn:
closes #15 

Adds tests for all of the existing modules, attempting to exercise all outputs and input variables.

Only the s3 module is actually creating real infrastructure. The rest of the modules are mocking the cloudfoundry provider to speed test runs significantly and enable testing modules that require either multiple spaces or more memory than we can dedicate to CI runs.


.terraform-version
```diff
@@ -0,0 +1 @@
+1.8.5
```
mogul, line 1: This may be a big jump for a lot of existing users. It looks like [the new testing features landed in 1.6](https://developer.hashicorp.com/terraform/language/tests#:~:text=Note%3A%20This%20testing%20framework%20is%20available%20in%20Terraform%20v1.6.0%20and%20later.). (You'd need 1.7 if you were using mocks or passing outputs from one provider into another, but you're not.) So maybe just ask for 1.6.6 here to better reflect the real minimum?

mogul (commented):
Good start! Looking forward to seeing how this develops.

asteel-gsa:
Very cool! This looks awesome!

.terraform-version
rahearn: I think 1.8.5 is fine here, since that's just the requirement for using terraform from this directory - like when running tests, and not the requirement for use of the modules. That is still specified as `~> 1.0` for now

anna-m-gsa (approved):

.terraform-version
rahearn: also, now we're using mocks

mogul: I saw, awesome! Definitely gonna need a download from you when I get back to the module(s) I'm working on.

clamav/tests/creation.tftest.hcl
```diff
@@ -0,0 +1,121 @@
...
+    error_message = "Routing policy allows traffic from the source app"
+  }
+
+  assert {
+    condition     = [for policy in cloudfoundry_network_policy.clamav_routing.policy : policy.destination_app][0] == cloudfoundry_app.clamav_api.id
+    error_message = "Routing policy allows traffic to the clamav app"
+  }
+
+  assert {
+    condition     = [for policy in cloudfoundry_network_policy.clamav_routing.policy : policy.port][0] == "61443"
```
rahearn, line 89: This line (along with others with the same shape) could also be written as

```suggestion
    condition     = [for policy in cloudfoundry_network_policy.clamav_routing.policy : policy.port] == ["61443"]
```
or
```suggestion
    condition     = tolist(cloudfoundry_network_policy.clamav_routing.policy)[0].port == "61443"
```

Does anyone have feelings about which is best?

rahearn: FWIW, I'm leaning towards changing everything to the first suggestion version, since that tests not just that the first entry has the expected value, but that there is only the one expected value. The other two versions would still pass with extra policies beyond the tested one.

rahearn: updated in 83b73f7

pauldoomgov (approved):
One non-blocking suggestion. This is really great! Making Terraform testable with native capabilities is so much slicker than alternatives!

pauldoomgov (approved):
LGTM!

.terraform-version
mogul: [OpenTofu 1.8.0-alpha1 adds support for mocks!](https://opentofu.org/blog/help-us-test-opentofu-1-8-0-alpha1/#resource-overrides-in-tofu-test)

mogul: I'm wrong, that's support for resource overrides, which was [a different Terraform 1.7 addition](https://www.hashicorp.com/blog/terraform-1-7-adds-test-mocking-and-config-driven-remove#:~:text=%7D-,Overrides,-In%20addition%20to). Hopefully mocking providers is coming!🤞

PR #57, opened 2024-11-21
Migrate provider from cloudfoundry-community to cloudfoundry

### LINKED ISSUE #56: Update to new cloudfoundry provider
## Story

As a ***devops engineer***,
I would like to ***use a better supported terraform provider for cloudfoundry***,
So that I can ***be confident that everything works properly and will continue to work as api v2 goes away***

## Solution

```[tasklist]
### Tasks
- [x] Utilize https://registry.terraform.io/providers/cloudfoundry/cloudfoundry/latest/docs provider
- [x] Determine if this is a breaking change and rev the semver appropriately
- [x] Add UPGRADING.md file with information on mix-and-match new and old module versions & how to upgrade an existing module
- [x] Add a v1 branch to enable future bug fixes on that line.
```

## Acceptance Criteria

- [ ] READMEs have been updated for appropriate version numbers
- [ ] space ssh and ASG things work properly

rahearn:
# 🎫 Addresses issue: #56

closes #56 

## 🛠 Summary of changes

* updates all resources to use the new cloudfoundry/cloudfoundry provider _except_ clamav and egress_proxy both still have references to the community provider to use `cloudfoundry_network_policy`
* lots of breaking changes with input variables, getting rid of unneeded `cf_org_name` and changing from `cf_space_name` to `cf_space_id` or `cf_space` (entire space resource) to help terraform properly sequence creation
* I had issues trying to mock out `cloudfoundry_service_plans` data so I switched to using more real provider in tests and less mocking, and using `override_resource` to prevent creating expensive infrastructure
* There is not currently an upgrade path for apps using the existing v1.x modules. I plan on adding an `UPGRADING.md` file with tips, but I also think most apps will stick on the version they started with for the most part.

## 📜 Testing Plan

While this is in draft, I used it in a new application to verify behavior, and attempt to update an existing app to generate content for the UPGRADING doc.

It's been used for both https://github.com/GSA-TTS/continuous-assessment-poc/pull/66 and https://github.com/GSA-TTS/gitlab-runner-cloudgov/pull/62


anna-m-gsa (dismissed):

database/main.tf
```diff
@@ -1,16 +1,17 @@
...
 
 resource "cloudfoundry_service_instance" "rds" {
   name         = var.name
-  space        = data.cloudfoundry_space.space.id
-  service_plan = data.cloudfoundry_service.rds.service_plans[var.rds_plan_name]
-  tags         = var.tags
-  json_params  = var.json_params
+  space        = var.cf_space_id
+  type         = "managed"
+  service_plan = data.cloudfoundry_service_plans.rds.service_plans.0.id
```
mogul, line 14: Shouldn't `var.rds_plan_name` be used to index into `service_plans` here?

README.md
```diff
@@ -129,10 +129,11 @@ Creates a new cloud.gov space, such as when creating an egress space, and output
 
 ```
 module "egress_space" {
-  source = "github.com/GSA-TTS/terraform-cloudgov//cg_space?ref=v1.1.0"
+  source = "github.com/GSA-TTS/terraform-cloudgov//cg_space?ref=v2.0.0"
```
mogul: Should this be `v2.0.0beta1` to match the other modules?

domain/main.tf
```diff
@@ -1,53 +1,55 @@
...
-  endpoint     = (var.host_name != null ? "${var.host_name}.${var.domain_name}" : var.domain_name)
-  target_apps  = (var.app_name_or_id != null ? [var.app_name_or_id] : var.app_names_or_ids)
-  service_name = (var.name == "" ? "${data.cloudfoundry_app.app[local.target_apps[0]].name}-${var.domain_name}" : var.name)
-}
-
-data "cloudfoundry_space" "space" {
-  org_name = var.cf_org_name
-  name     = var.cf_space_name
+  service_name  = (var.name == "" ? "${var.app_names[0]}-${var.domain_name}" : var.name)
+  tags          = setunion(["terraform-cloudgov"], var.tags)
```
mogul, line 3:
```suggestion
  tags          = setunion(["terraform-cloudgov-managed"], var.tags)
```

domain/main.tf
```diff
@@ -1,53 +1,55 @@
...
 }
 
 resource "cloudfoundry_service_instance" "external_domain_instance" {
   name         = local.service_name
-  space        = data.cloudfoundry_space.space.id
-  service_plan = data.cloudfoundry_service.external_domain.service_plans[var.cdn_plan_name]
-  json_params  = "{\"domains\": \"${local.endpoint}\"}"
-  tags         = var.tags
+  space        = var.cf_space.id
+  service_plan = data.cloudfoundry_service_plans.external_domain.service_plans.0.id
```
mogul, line 51: Shouldn't `var.cdn_plan_name` be used to index into `service_plans` here?

redis/main.tf
```diff
@@ -1,16 +1,17 @@
-data "cloudfoundry_space" "space" {
-  org_name = var.cf_org_name
-  name     = var.cf_space_name
+locals {
+  tags = setunion(["terraform-cloudgov"], var.tags)
```
mogul, line 2:
```suggestion
  tags = setunion(["terraform-cloudgov-managed"], var.tags)

```

redis/main.tf
```diff
@@ -1,16 +1,17 @@
...
 
 resource "cloudfoundry_service_instance" "redis" {
   name         = var.name
-  space        = data.cloudfoundry_space.space.id
-  service_plan = data.cloudfoundry_service.redis.service_plans[var.redis_plan_name]
-  tags         = var.tags
-  json_params  = var.json_params
+  space        = var.cf_space_id
+  type         = "managed"
+  service_plan = data.cloudfoundry_service_plans.redis.service_plans.0.id
```
mogul, line 14: Shouldn't `var.redis_plan_name` be used as an index into `service_plans` here?

s3/main.tf
```diff
@@ -1,16 +1,17 @@
-data "cloudfoundry_space" "space" {
-  org_name = var.cf_org_name
-  name     = var.cf_space_name
+locals {
+  tags = setunion(["terraform-cloudgov"], var.tags)
```
mogul, line 2:
```suggestion
  tags = setunion(["terraform-cloudgov-managed"], var.tags)
```

s3/main.tf
```diff
@@ -1,16 +1,17 @@
...
 
 resource "cloudfoundry_service_instance" "bucket" {
   name         = var.name
-  space        = data.cloudfoundry_space.space.id
-  service_plan = data.cloudfoundry_service.s3.service_plans[var.s3_plan_name]
-  tags         = var.tags
-  json_params  = var.json_params
+  space        = var.cf_space_id
+  type         = "managed"
+  service_plan = data.cloudfoundry_service_plans.s3.service_plans.0.id
```
mogul, line 14: Shouldn't `var.s3_plan_name` be used to index into `service_plans` here?

s3/tests/creation.tftest.hcl
```diff
@@ -23,7 +23,7 @@ run "test_bucket_creation" {
   }
 
   assert {
-    condition     = cloudfoundry_service_instance.bucket.service_plan == data.cloudfoundry_service.s3.service_plans[var.s3_plan_name]
+    condition     = cloudfoundry_service_instance.bucket.service_plan == data.cloudfoundry_service_plans.s3.service_plans.0.id
```
mogul, line 26: I think this passing with `basic-sandbox` is a coincidence that depends on that being the first plan listed, isn't it?

README.md
```diff
@@ -156,10 +156,10 @@ Prerequities:
...
-  cf_org_name     = local.cf_org_name
-  cf_egress_space = data.cloudfoundry_space.egress_space
-  cf_client_space = data.cloudfoundry_space.app_space
-  name            = "egress-proxy"
+  cf_org_name      = local.cf_org_name
+  cf_egress_space  = data.cloudfoundry_space.egress_space
+  cf_client_spaces = {(data.cloudfoundry_space.app_space.name) = data.cloudfoundy_space.app_space.id}
+  name             = "egress-proxy"
   allowlist = {
     "source_app_name" = ["host.com:443", "otherhost.com:443"]
```
mogul, line 164: Since we're now allowing multiple client spaces, the app names specified as keys may not be unique. We should probably set a new naming convention that will keep those keys useful for distinguishing between clients later while still being easy to read, eg:
```suggestion
    "source_space_name/source_app_name" = ["host.com:443", "otherhost.com:443"]
```

In this case, maybe we should get the list of space names from those indices and just drop the `cf_client_space(s)` variable altogether as redundant.

clamav/variables.tf
```diff
@@ -3,12 +3,15 @@ variable "cf_org_name" {
...
+variable "cf_space" {
+  type = object({
+    id   = string
+    name = string
+  })
+  description = "cloud.gov space resource"
 }
 
-variable "app_name_or_id" {
+variable "app_name" {
```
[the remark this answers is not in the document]
mogul: I see that you removed the network policy creation code for the egress module in a later commit; that makes me more confident that we should remove this behavior!

mogul (changes requested):
Most of these are optional, but I think the service-plan indexing ones are important.

cg_space/main.tf
```diff
@@ -3,46 +3,30 @@ data "cloudfoundry_org" "org" {
...
-  name     = each.key
-  org_id   = data.cloudfoundry_org.org.id
-}
-
-data "cloudfoundry_user" "deployers" {
-  for_each = var.deployers
-  name     = each.key
-  org_id   = data.cloudfoundry_org.org.id
+locals {
+  manager_names   = setunion(var.managers, var.deployers)
```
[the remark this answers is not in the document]
rahearn: The reason both the old and new code declares a "deployer" to have both developer and manager roles is because manager is required in order to set the egress security group on the space.

cg_space/tests/creation.tftest.hcl
```diff
@@ -39,54 +23,59 @@ run "test_space_creation" {
...
-    developers = ["user.developer@gsa.gov"]
-    deployers  = ["user.manager@gsa.gov"]
+    developers = ["paul.hirsch@gsa.gov"]
+    deployers  = ["ryan.ahearn@gsa.gov"]
   }
 
   assert {
-    condition     = cloudfoundry_space_users.space_permissions.managers == toset(["1e5143a4-aa47-483c-8352-557988d5cc7a"])
+    condition     = keys(cloudfoundry_space_role.managers) == ["ryan.ahearn@gsa.gov"]
     error_message = "Should be able to set Space Managers via var.deployers"
```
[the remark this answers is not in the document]
rahearn: Resolving to track this discussion on the other thread

clamav/variables.tf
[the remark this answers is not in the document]
rahearn: Yeah, I'll be revisiting this as I further use this branch. Definitely had to remove from egress for circular dependency reasons.

README.md
rahearn: Agreed on revisiting this. I am avoiding just passing space names around because you end up with nasty race conditions where terraform thinks it can operate on a space but that space does not yet exist and you have to add a ton of `depends_on = []` to your root module

cg_space/main.tf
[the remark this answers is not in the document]
mogul: Right, but do you want the deployer to do that? It seems like this module should also manage [a `cloudfoundry_space_asgs` resource](https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs/resources/space_asgs) with the desired ASGs passed in a variable.

database/main.tf
mogul: Never mind, I just learned that supplying `name = var.rds_plan_name` to `cloudfoundry_service_plans` will filter the result to just that one service plan, so there will only be one result and `.0.` is always right!

domain/main.tf
mogul: Never mind, I just learned that supplying `name = var.cdn_plan_name` to `cloudfoundry_service_plans` will filter the result to just that one service plan, so there will only be one result and `.0.` is always right!

redis/main.tf
mogul: Never mind, I just learned that supplying `name = var.redis_plan_name` to `cloudfoundry_service_plans` will filter the result to just that one service plan, so there will only be one result and `.0.` is always right!

s3/main.tf
mogul: Never mind, I just learned that supplying `name = var.s3_plan_name` to `cloudfoundry_service_plans` will filter the result to just that one service plan, so there will only be one result and `.0.` is always right!

README.md
mogul: It looks like it's only used for [creating an UPSI with client credentials in each space where clients exist](https://github.com/GSA-TTS/terraform-cloudgov/blob/a11cbd14c8fe694778dcf734c135fa1798728298/egress_proxy/main.tf#L76-L89). I think a better solution is for the module to do less: just return the credential JSON in an output variable, and make it the caller's job to decide the best way for that information to be used. (They may well choose to create an UPSI with that content in each client space, or they may grab values out and swizzle them into some other form, pass info via env vars, etc.)

mogul: (That has the side benefit of resolving dependency cycles, because if the UPSIs are created outside the module, Terraform has much better control of implicit dependencies with the spaces.)

mogul:
> Most of these are optional, but I think the service-plan indexing ones are important.

OK, turns out those were _not_ important because `cloudfoundry_service_plans` filters the results; I understand that now and resolved those.

cg_space/main.tf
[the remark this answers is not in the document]
rahearn: Agreed, but the community asgs resource is totally broken and the official one exists only as a PR right now (though as soon as it's released I'll be adding it to the cg_space module so we can setup the space, user roles, and security groups in one go.

egress_proxy/main.tf
```diff
@@ -61,31 +52,12 @@ resource "cloudfoundry_app" "egress_app" {
...
   https_proxy = "https://${random_uuid.username.result}:${random_password.password.result}@${local.egress_route}:61443"
   http_proxy  = "http://${random_uuid.username.result}:${random_password.password.result}@${local.egress_route}:8080"
   domain      = local.egress_route
   username    = random_uuid.username.result
   password    = random_password.password.result
-  protocol    = "https"
-  port        = 61443
-}
-
-resource "cloudfoundry_service_instance" "credentials" {
```
mogul: Since we're no longer creating the UPSI, I think we should make it as easy as possible for callers to create the UPSI (and maybe show them how in the `README.md`). To that end, should we add the `jsonencode()`'d block below to a new output, eg `json_credentials`?

egress_proxy/outputs.tf
```diff
@@ -3,12 +3,21 @@ output "https_proxy" {
...
+output "http_proxy" {
+  value     = local.http_proxy
+  sensitive = true
+}
+
 output "domain" {
   value = local.domain
 }
 
-output "port" {
```
mogul: Should we leave `port` with the same value as `https_port` for back-compat/migration purposes?

mogul (dismissed):

UPGRADING.md
```diff
@@ -0,0 +1,119 @@
...
+```
+resource "cloudfoundry_service_instance" "egress_proxy_credentials" {
+  name        = "egress-proxy-credentials"
+  space       = module.app_space.space_id
+  type        = "user-provided"
+  credentials = module.egress_proxy.json_credentials
+}
+```
+
+### Clamav
```
mogul, line 101:
```suggestion
### ClamAV
```

UPGRADING.md
```diff
@@ -0,0 +1,119 @@
...
+
+Clamav no longer sets up network policies between the clamav app and client apps. It is the developer's responsibility to set this up to better handle circular dependencies between the various apps.
```
mogul, line 103:
```suggestion
ClamAV no longer sets up network policies between the clamav app and client apps. It is the developer's responsibility to set this up to better handle circular dependencies between the various apps.
```

UPGRADING.md
```diff
@@ -0,0 +1,119 @@
...
+
```
mogul, line 104: We should probably duplicate [these lines](https://github.com/GSA-TTS/terraform-cloudgov/pull/57/files/169e6ed7c05baff02d750cad1e54428568d190c9..35a63aa2d5c4d2e6d1c8ddee9f75e250813958ef#diff-f411a5009131db5a6c7242b49ca78488f2d274ef9e5e5f4de9262a42108a5386R77) here in case the reader is not using the egress proxy and skipped past the section above.

UPGRADING.md
```diff
@@ -0,0 +1,119 @@
...
+### cg_space
+
+Migrating the space roles is a little trickier. Here is a path forward:
```
mogul, line 107: I'm not really clear on what's happening in these steps. Can you summarize what's changed as you did for the other modules?

mogul (changes requested):

PR #58, opened 2024-12-08
Add spiffworkflow module

mogul:
## 🛠 Summary of changes

* Adds a module for deploying [SpiffWorkflow](https://spiffworkflow.org) on [cloud.gov](https://cloud.gov)

Still in draft form; needs

* ~A GitHub action to scan and republish the upstream images in this repo~
  * As a bonus, adds a default image for clamav
* ~Use the hash rather than the tag for the supplied imageref~
* ~Verify that we gsa_default_models will get pulled~
* ~Random routes by default to minimize collisions~
* ~More testing~
* ~actual hcltest files!~


asteel-gsa:
Note for tomorrow:
- Images are built and stored in this repo
- Process models aren't showing up. My _hunch_ is that it needs an ssh key to use git to connect to that repo, and do the pull for the process models. It will have to be a required variable if that is the case, as there wouldn't be a point to deploying it without the process models. 
- Find out if we can use a sha instead of tags for the images (needs some research)
- Write tests

spiffworkflow/main.tf
```diff
@@ -0,0 +1,219 @@
...
+# }
+
+# resource "cloudfoundry_service_instance" "database" {
+#   name         = "${var.app_prefix}-database"
+#   space        = data.cloudfoundry_space.space.id
+#   type         = "managed"
+#   service_plan = data.cloudfoundry_service_plans.rds.service_plans.0.id
+#   tags         = local.tags
+#   parameters   = var.rds_json_params
+# }
```
asteel-gsa, line 49: tests don't seem to pass unless this is commented out. any support would be welcome on this front

spiffworkflow/main.tf
```diff
@@ -0,0 +1,219 @@
...
+    # curl https://spiffworkflow((slug))-connector.apps.internal:61443/liveness
+
+    /app/bin/clone_process_models
+    /app/bin/boot_server_in_docker
+    COMMAND
+  health_check_type          = "http"
+  health_check_http_endpoint = "/api/v1.0/status"
+  service_bindings = [
+    # { service_instance = cloudfoundry_service_instance.database.name }
+    { service_instance = var.database_service_instance_id }
```
asteel-gsa, line 101: it might just be better long-term to have the team be responsible for creating the database, and then passing the id via `database_service_instance_id = data.cloudfoundry_service.spiff_database.id`

spiffworkflow/main.tf
mogul: @rahearn can you help here? I'm unfamiliar with Terraform tests.

rahearn: I opened https://github.com/GSA-TTS/terraform-cloudgov/pull/66 with how I've gotten around issues of being completely unable to mock out `data.cloudfoundry_service_plans` in the other modules.

spiffworkflow/main.tf
rahearn: That's another decent way to get aroun the testing issue, especially if we expect they'd need the DB independent of this module.

spiffworkflow/main.tf
asteel-gsa: So, I stumbled on using the actual provider as well, I was hoping to get away with overrides and mocks, not using the actual provider, as one flaw I am seeing is it requires a clean environment. I was able to get your branch completed @rahearn, so long as I deleted everything prior.

Now, this is one way we can do it, but I am starting to think, we just update the readme.md to consume a service id for the database, make it a required var, and effectively just do a

```tf
module "SpiffWorkflow" {
  source                       = "github.com/gsa-tts/terraform-cloudgov//clamav?ref=v2.2.0"
  cf_org_name                  = var.cf_org_name
  cf_space_name                = var.cf_space_name
  process_models_ssh_key       = var.process_models_ssh_key
  database_service_instance_id = data.cloudfoundry_service_instance.spiff_db.id
  tags                         = []
  depends_on                   = [module.spiff_database]
}

module "spiff_database" {
  source        = "github.com/gsa-tts/terraform-cloudgov//database?ref=v2.2.0"
  cf_space_id   = var.cf_space_id
  name          = "spiffworkflow-db"
  tags          = ["rds"]
  rds_plan_name = var.database_plan
  json_params   = var.json_params
}

data "cloudfoundry_service_instance" "spiff_db" {
  name  = "spiffworkflow-db"
  space = var.cf_space_id
}
```

That way we can keep the mock tests, resources can already exist in the target org/space, and then we offload the database creation from this module, but effectively make it a dependency. Happy to go with whatever you and @mogul deem acceptable. Either way works for me. I did get #66 running and tests passing though. So happy to do whichever.

rahearn: I'm fine with making db an external dependency.

I will note that the official `cloudfoundry_app` resource requires the service name, not the service ID for its bindings (which is super annoying, because it means you have to use `depends_on` to manually order creation)

asteel-gsa: I can confirm that I was able to get this fully deployed with the following:

```tf
data "cloudfoundry_org" "org" {
  name = var.cf_org_name
}

data "cloudfoundry_space" "space" {
  name = var.cf_space_name
  org  = data.cloudfoundry_org.org.id
}

module "SpiffWorkflow" {
  source                       = "../spiffworkflow"
  cf_org_name                  = var.cf_org_name
  cf_space_name                = var.cf_space_name
  process_models_ssh_key       = var.process_models_ssh_key
  database_service_instance_name = "spiffworkflow-db"
  tags                         = ["SpiffWorkflow"]
  depends_on                   = [ module.Database ]
}

module "Database" {
  source        = "github.com/gsa-tts/terraform-cloudgov//database?ref=v2.2.0"
  cf_space_id   = data.cloudfoundry_space.space.id
  name          = "spiffworkflow-db"
  tags          = ["rds", "SpiffWorkflow"]
  rds_plan_name = "small-psql"
}
```

Pushing changes now

asteel-gsa:
Alright, this is about as ready as it will ever be.. _I think_ 🤣 @mogul

asteel-gsa:
Think this is ready for your review @rahearn 

CC @mogul 
image

.github/workflows/pull-and-publish-images.yml
```diff
@@ -0,0 +1,82 @@
...
+          ignore-unfixed: true
+
+      - name: Upload Trivy scan results to GitHub Security tab for Third Party Images
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: 'trivy-results.sarif'
+
+      # - name: Scan Image
+      #   run: docker run aquasec/trivy:latest image --db-repository public.ecr.aws/aquasecurity/trivy-db,ghcr.io/aquasecurity/trivy-db --java-db-repository public.ecr.aws/aquasecurity/trivy-java-db,ghcr.io/aquasecurity/trivy-java-db --timeout 5m --scanners vuln --exit-code 1 --severity CRITICAL,HIGH  ${{ matrix.image.name }}
+
```
rahearn, line 67:
```suggestion
```

.github/workflows/pull-and-publish-images.yml
```diff
@@ -0,0 +1,82 @@
...
+          exit-code: 0
+          severity: 'CRITICAL,HIGH'
+          scanners: 'vuln'
+          timeout: 15m0s
+          ignore-unfixed: true
+
+      - name: Upload Trivy scan results to GitHub Security tab for Third Party Images
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: 'trivy-results.sarif'
```
rahearn, line 63: [question] is this going to work as expected in the matrix setup? I'm concerned that each image will overwrite findings in other images

.github/workflows/pull-and-publish-images.yml
```diff
@@ -0,0 +1,82 @@
...
+          - name: ghcr.io/gsa-tts/clamav-rest/clamav:latest
+            short-name: clamav-proxy-support
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v3
```
rahearn, line 35: [question] do we need the buildx setup? Seems like we're just using vanilla docker commands for the rest of this action

.github/workflows/update-trivy-cache.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Download and extract the Java DB
+        run: |
+          mkdir -p $GITHUB_WORKSPACE/.cache/trivy/java-db
+          oras pull ghcr.io/aquasecurity/trivy-java-db:1
+          tar -xzf javadb.tar.gz -C $GITHUB_WORKSPACE/.cache/trivy/java-db
+          rm javadb.tar.gz
+
+      - name: Cache DBs
+        uses: actions/cache/save@v4
```
rahearn, line 35: [question] Does the pull/publish workflow need an explicit cache restore, or does the trivy-action do that for us?

spiffworkflow/providers.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = "~> 1.0"
+  required_providers {
+    cloudfoundry = {
+      source  = "cloudfoundry/cloudfoundry"
+      version = ">=1.1.0"
```
rahearn, line 6: [suggestion] there have been important bug fixes in recent versions, so I'd start this at the most recent available today:

```suggestion
      version = ">=1.3.0"
```

spiffworkflow/variables.tf
```diff
@@ -0,0 +1,126 @@
...
+}
+
+variable "process_models_repository" {
+  type        = string
+  description = "git repository with process models (for read-write, use SSH-style 'git@github.com:...' and supply your ssh_key)"
+}
+
+variable "process_models_ssh_key" {
+  type        = string
+  description = "private SSH key (only needed for read-write to SSH-style 'git@github.com:...' repositories)"
```
rahearn, line 50: [question] should this have a `default = null` if it's only needed for ssh-style repos? How would you configure a read-only repo?

spiffworkflow/main.tf
```diff
@@ -0,0 +1,195 @@
...
+  connector_route = "${local.route_prefix}-connector.apps.internal"
+  frontend_route  = "${local.route_prefix}.app.cloud.gov"
+
+  username = random_uuid.username.result
+  password = random_password.password.result
+
+  # backend_url   = "https://${local.username}:${local.password}@${local.backend_route}"
+  # connector_url = "https://${local.connector_route}:61443"
+  # frontend_url  = "https://${local.username}:${local.password}@${local.frontend_route}"
+
```
rahearn, line 14:
```suggestion
```

rahearn (commented):
Some feedback - I haven't yet tried to run the module

.github/workflows/pull-and-publish-images.yml
asteel-gsa: I haven't noticed anything weird in the FAC repo, though it does bring up a thought.. 
We maintain 3 images on the fac, postgrest, clamav, and a container build. The container and clamav both show vulnerabilities, and postgrest doesn't publish if it detects. So we don't see any postgrest vuln in the security tab.

I know there are some vulns, so I can run it from this branch and publish to the security tab, and, if they overwrite, I can try to matrix the files to `$({ matrix.image.short-name })-results.sarif`

.github/workflows/pull-and-publish-images.yml
asteel-gsa: This was something that I did a while back, I think it was in relation to the old method I used to build the containers, so I can pull it out, can;t exactly remember why I used buildx other than trouble shooting something a while back, and buildx seemed to work a while back. Ill test it

.github/workflows/update-trivy-cache.yml
asteel-gsa: Yes, this is a workaround for a recent Trivy issue that relates to ratelimiting on downloading the vuln dbs on the fly. https://github.com/aquasecurity/trivy-action?tab=readme-ov-file#cache https://github.com/aquasecurity/trivy-action/issues/389

We noticed a high amount of failures due to ratelimiting on the FAC, and since I put the dbs in the github cache, we have not encountered this ratelimiting issue since.

spiffworkflow/variables.tf
asteel-gsa: The ssh key is responsible for not only cloning the repo down to load the "example" process models, but also commiting back to it with changes. For example, editing a .bpmn directly on the application, directly commits it back into the target branch, to allow a PR to merge back into main for a permanent update to said models.

Not entirely sure this supports read only, as someone with admin rights on the app can directly edit the bpmn, and will throw an error when saving the model if it cannot commit back to the repo. 

To that point, we can't specify a null value for the var, as a ssh key is a requirement for loading the initial models from https://github.com/gsa-TTS/gsa-process-models via git ssh. It will need to have a value for full functionality.

.github/workflows/pull-and-publish-images.yml
asteel-gsa: Should be all set regardless, as each matrix runs the scan and publish results on their own runner, but just to be safe, added a unique name for each sarif file

clamav/variables.tf
```diff
@@ -16,6 +16,8 @@ variable "name" {
 variable "clamav_image" {
   type        = string
   description = "Docker image to deploy the clamav api app"
+  # Uses the https://github.com/GSA-TTS/clamav-rest clamav image, which is maintained by GSA-TTS, and supports a proxy.
+  default = "ghcr.io/gsa-tts/terraform-cloudgov/clamav-proxy-support:latest"
```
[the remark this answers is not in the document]
asteel-gsa: of course! FAC forked the repo and made some small changes, notably [this](https://github.com/GSA-TTS/clamav-rest/blob/8ac7574fc6e132c954132cabae7340595d9967fa/entrypoint.sh#L17-L30) and [this](https://github.com/GSA-TTS/clamav-rest/blob/8ac7574fc6e132c954132cabae7340595d9967fa/entrypoint.sh#L52-L65), which this repos image is pulled from, so the default for the proxy module is now a clamav that will guarantee proxy support, should `$PROXY_SERVER` be set.

spiffworkflow/variables.tf
```diff
@@ -0,0 +1,126 @@
...
+
+variable "tags" {
+  description = "A list of tags to add to the module's resource"
+  type        = set(string)
+  default     = []
+}
+
+variable "process_models_repository" {
+  type        = string
+  description = "git repository with process models (for read-write, use SSH-style 'git@github.com:...' and supply your ssh_key)"
```
rahearn, line 45: [suggestion] to avoid confusion about whether read-write is optional:

```suggestion
  description = "git repository with process models (use SSH-style 'git@github.com:...')"
```

spiffworkflow/variables.tf
rahearn: [suggestion] to avoid confusion about whether read-write is optional:

```suggestion
  description = "private SSH key with read/write access to var.process_models_repository repo"
```

asteel-gsa: Done!

spiffworkflow/variables.tf
```diff
@@ -0,0 +1,126 @@
...
+  default     = []
+}
+
+variable "process_models_repository" {
+  type        = string
+  description = "Git repository with process models (use SSH-style 'git@github.com:...')"
+}
+
+variable "process_models_ssh_key" {
+  type        = string
```
rahearn, line 49: [suggestion] mark this variable as sensitive

```suggestion
  type        = string
  sensitive   = true
```

rahearn (commented):
I tried deploying the module, but I'm getting this error when I try to visit the frontend:

```
{
"error_code":"internal_server_error",
"message":"OpenIdConnectionError: Cannot connect to given open id url: https://spiffworkflow-harmless-newt.app.cloud.gov/api/openid/.well-known/openid-configuration",
"status_code":500}
```

asteel-gsa:
> I tried deploying the module, but I'm getting this error when I try to visit the frontend:
> 
> ```
> {
> "error_code":"internal_server_error",
> "message":"OpenIdConnectionError: Cannot connect to given open id url: https://spiffworkflow-harmless-newt.app.cloud.gov/api/openid/.well-known/openid-configuration",
> "status_code":500}
> ```

Do you by chance not have this? https://github.com/GSA-TTS/spiff-arena/pull/1#issuecomment-2045849524

asteel-gsa:
Added network policy into the module and update readme per convo @rahearn

rahearn (approved):
🚢

README.md
```diff
@@ -210,9 +210,9 @@ module "drupal" {
...
 
 **NOTE:**
-You must have a valid git key pairing. Generate with ssh-keygen -t rsa -b 4096 -C "my-git@email", and add the public key to **https://github.com/settings/keys**. var.process_models_ssh_key is the private key. When you store process_models_ssh_key in a .tfvars, ensure that the file format of the .tfvars file is in "LF" End Of Line Sequence. **This key is a profile level SSH key, and does not appear to work at the repo level**
-
-Ensure that `cf bind-security-group public_networks_egress ORGNAME --lifecycle running --space SPACENAME` has been set for your target space.
+1. You must have a valid git key pairing. Generate with ssh-keygen -t rsa -b 4096 -C "my-git@email", and add the public key to **https://github.com/settings/keys**. var.process_models_ssh_key is the private key. When you store process_models_ssh_key in a .tfvars, ensure that the file format of the .tfvars file is in "LF" End Of Line Sequence. **This key is a profile level SSH key, and does not appear to work at the repo level**
+2. Ensure that `cf bind-security-group public_networks_egress ORGNAME --lifecycle running --space SPACENAME` has been set for your target space.
```
mogul, line 214: Is this needed for the connector...? Or maybe for pulling/pushing processes from/to GitHub? Regardless, it seems like it should be accepting an egress proxy, and documenting what URLs should be allowed by that egress proxy for it to work properly.

README.md
```diff
@@ -210,9 +210,9 @@ module "drupal" {
...
+3. Ensure that your space has the `public_networks_egress`security group.
```
mogul, line 215: Is this redundant with the above line?

README.md
mogul: Note we can take this up in another PR just to get this merged. However it seems like it's going to be a necessary change before this can be used in a production context.

rahearn: Agreed about it being a necessary step for production use - but I also think there's quite a ways to go before it's production ready and I didn't want to hold this up until then. That's why there's the disclaimer in the readme now

rahearn (approved):

README.md
asteel-gsa: Yea, I think this was simply because non of us were running it behind a proxy, and I pulled that NSG from a comment you and I did on the long-lived branch we have, which did resolve the front end for Ryan!

PR #60, opened 2024-12-20
Fix egress proxy route bug

rahearn:
# 🎫 Addresses issue: no issue

## 🛠 Summary of changes

* validates that the default route name for the egress proxy starts with a letter
* adds the optional ability to manually specify what the route hostname should be
* updates some obsolete README content from the v2 transition

### DIFF
--- a/egress_proxy/tests/creation.tftest.hcl
+++ b/egress_proxy/tests/creation.tftest.hcl
@@ -45,4 +45,30 @@ run "test_proxy_creation" {
     condition     = output.http_port == 8080
     error_message = "http_port reports port 8080 for plaintext"
   }
+
+}
+
+run "test_specific_hostname_bug" {
+  variables {
+    cf_org_name = "gsa-tts-devtools-prototyping"
+    cf_egress_space = {
+      id   = "169c6e21-2513-43f7-bbff-80cc5e456882"
+      name = "rca-tfm-stage-egress"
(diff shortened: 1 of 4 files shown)


egress_proxy/tests/creation.tftest.hcl
```diff
@@ -45,4 +45,30 @@ run "test_proxy_creation" {
...
+}
+
+run "test_specific_hostname_bug" {
+  variables {
+    cf_org_name = "gsa-tts-devtools-prototyping"
+    cf_egress_space = {
+      id   = "169c6e21-2513-43f7-bbff-80cc5e456882"
+      name = "rca-tfm-stage-egress"
+    }
+    name = "egress-proxy-staging"
```
rahearn, line 58: this specific set of org/space/app name resulted in a `-devtools...` route name

btylerburton (approved):

pauldoomgov (approved):
I wish apps.internal was scoped to the org level, but alas... I think squishing and keeping the last 63 chars is the least bad option, and thanks for the test! LGTM!

egress_proxy/tests/creation.tftest.hcl
rahearn: well - it'll keep the internal hyphens for some level of readability (not sure if that's important or not)

PR #65, opened 2025-02-26
Add basic drupal app module

rahearn:
# 🎫 Addresses issue: no issue

current state:

This module works with my fork of beta.weather.gov's drupal app, to have a working sandbox with only:

1. `terraform apply`
2. `cf run-task "scripts/import-content.sh"...`

marking this as draft until the code bits from my fork are at least documented in an easy-to-follow manner.

## 🛠 Summary of changes

* adds a single module to stand up a drupal app and supporting services
* TODO: fully document the setup steps to go from new project -> using this module, including:
  * install s3fs module and configure
  * .bp-config/options.json, scripts/post-deploy.sh, and scripts/cf-shell.sh customizations from beta.weather.gov
  * custom settings file to read in from `VCAP_SERVICES`


drupal/temp.creds
```diff
@@ -0,0 +1,5 @@
+# generated with /Users/ryancahearn/.bin/create_service_account.sh -s terraform-cloudgov-ci-tests -u terraform-test-ci-runner -r space-deployer -o gsa-tts-devtools-prototyping
+# revoke with /Users/ryancahearn/.bin/destroy_service_account.sh -s terraform-cloudgov-ci-tests -u terraform-test-ci-runner -o gsa-tts-devtools-prototyping
+
+cf_user = "d2c517b2-b9dd-44f6-b286-11d0cd111aba"
+cf_password = "65+fcaP/gOkU8aW:bcqMw62ggWykPlLG"
```
rahearn, line 5: oops - but this is fine since the user has already been deleted

README.md
```diff
@@ -173,6 +173,34 @@ module "egress_proxy" {
...
 See <UPGRADING.md> for an example of how to set up network policies and credential stores to enable your client app to reach the proxy.
 
+### drupal
+
+> [!WARNING]
+> This module is in active development and may change
+
+Creates and configures a Drupal application, and basic supporting services needed.
+
+There are also code changes required to take advantage of this module. Read more at TKTK
```
zjrgov, line 183: Should we log an issue for the read more or something?

zjrgov (approved):

README.md
rahearn: ah, no, I have a link now. thanks for catching it

zjrgov (approved):

drupal/outputs.tf
mogul, line 1: It feels like the ID for the creds UPS should also be in the outputs so that people can do things with the instance after it's up. For example a `null_resource` that runs `drush` to do further config stuff that needs to land in the DB (not filesystem-ephemeral stuff, obvs). I am totally making that up. But anyway, we could keep the option of composability with other stuff that would directly interact with Drupal open.

drupal/main.tf
```diff
@@ -0,0 +1,124 @@
+locals {
+  tags         = setunion(["terraform-cloudgov-managed"], var.tags)
+  s3_name      = "${var.name}-storage"
+  rds_name     = "${var.name}-database"
+  secrets_name = "${var.name}-secrets"
+  drupal_route = coalesce(var.route, "${var.name}.app.cloud.gov")
+}
+
+module "database" {
```
mogul, line 9: Maybe consider conditional creation/dependency injection for DB and S3 here in case the consumer of the module needs to do things with an existing resource not owned by this module. Did I hear correctly that dependency injection enabled better mocking for the SpiffWorkflow module, and if so, would that help here too?

mogul (commented):
Late to the party, but I thought you might still want my feedback for future issues/PRs.

drupal/main.tf
rahearn: Do you have an example of conditional creation working? I've wanted to do that in the past but I haven't been able to correctly set it up in a way that terraform was happy with. So I've been left with the choice of "does this module always handle this resource or does the user always have to pass it in" In _most_ cases I'd prefer "this module will create this resource unless this var is set"

mogul: I did it in the SMTP broker to implement [the "domain/no-domain" features](https://github.com/GSA-TTS/datagov-brokerpak-smtp/tree/main?tab=readme-ov-file#featurescomponents).

- [Based on the variables, determine if the module should handle required resource creation](https://github.com/GSA-TTS/datagov-brokerpak-smtp/blob/dc070fc5aaa709eb09a4a21ed10fba9ce5566946/terraform/provision/main.tf#L4)
- [Use that information with `count` to toggle the creation of those resources](https://github.com/GSA-TTS/datagov-brokerpak-smtp/blob/main/terraform/provision/verification.tf)

rahearn: ohhhh, you know what - I think it would work since we're using names and not ids. I believe what won't work is:

```
locals {
  db_id = (var.db_id == "" ? cloudfoundry_service_instance.db.0.id : var.db_id)
}
```

because the `cloudfoundry_service_instance.db` doesn't get created, but since we're just operating on the names we can avoid referencing the conditionally-created instance. Though now I'm worried about how to prevent the app from being created before the DB is ready

mogul: Yeah, I don't think you can make a `depends_on` refer to the exact ID of the optionally-created thing. It _might_ still work if you have the `depends_on` refer to `cloudfoundry_service_instance.db` array as a whole, since it'll still exist and just be zero-length. Not sure, you'd have to try it. 🤔

PR #66, opened 2025-03-05
Switch from mocked tests to real integration tests

rahearn:
# 🎫 Addresses issue: no issue

## 🛠 Summary of changes

* switch from using `mock_provider` to real provider because I can't figure out how to properly mock out `data.cloudfoundry_service_plans` in a way that makes terraform happy

this causes the test to take about 10 minutes, which is probably too long. We could [override the most expensive resources](https://github.com/GSA-TTS/terraform-cloudgov/blob/206ee1a196406a9f29a08aef57da88552400752b/database/tests/creation.tftest.hcl#L18-L23) to bring that back down while keeping the real provider in use.

### DIFF
--- a/spiffworkflow/main.tf
+++ b/spiffworkflow/main.tf
@@ -30,23 +30,23 @@ data "cloudfoundry_service_plans" "rds" {
   service_offering_name = "aws-rds"
 }
 
-# data "cloudfoundry_org" "org" {
-#   name = var.cf_org_name
-# }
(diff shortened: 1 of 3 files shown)


asteel-gsa:
Testing locally now! 1 moment.

mogul:
I agree, the DB should be injected into the module as a required parameter. That way the caller can determine its size, we can use a cheap or mocked one for testing, etc.

Issue #68, opened 2025-03-10
Add Application Module


asteel-gsa:
I set up everything for the module in this [branch](https://github.com/GSA-TTS/terraform-cloudgov/compare/main...add-application-module), however, the main way for me to validate it is to actually deploy the FAC application. Since this is a Django app, it does require an s3 service binding for the staticfiles to go to during bootup. Because of this, I do not want to hardcode in anything as a requirement, as that could vary from project to project.

Without the ability to perform a service binding after the fact, I am not exactly sure we can take this module in just yet. With that being said, so long as the app doesnt require service bindings to start up, or any service bindings for that matter, this should deploy no issue. In that event then the module can be used. 

However, at minimum, based on what we do at the FAC, we are looking, minimum, for most running apps:
```tf
  service_binding {
    service_instance = cloudfoundry_user_provided_service.clam.id
  }

  service_binding {
    service_instance = var.s3_id
  }

  service_binding {
    service_instance = var.db_id
  }

  service_binding {
    service_instance = var.https_proxy_creds_id
  }

  service_binding {
    service_instance = var.logshipper_id
  }
```

PR #69, opened 2025-03-10
Add application module

asteel-gsa:
# 🎫 Addresses issue: #68

TODO: 
- ~Finish hcl tests~
- ~cleanup~
- ~readme~

## 🛠 Summary of changes
Adds an application deployer module to the shared modules. 

Things to account for with this:
- ~I need to consider the removal of the service bindings, see #68, but, since neither `cloudfoundry/cloudfoundry` nor `cloudfoundry-community/cloudfoundry` have the ability to bind a service to an app via terraform, I don't know if the inclusion of a database and s3 bucket for all apps is the right call. Maybe, idk, tbd.~
  - It would still fall on the team to bind things like proxy, logshipper, etc
  - ~We need the bare minimum so that teams can do a `cf bind-service app_name service_instance` to bind their services after the fact.  I don't exactly like this approach, but it may be what has to be done, i.e a null resource that the team adds to just bind whatever they need.~
- I built this passing `DISABLE_COLLECTSTATIC = 1` environment var, as the "new" provider seems to not like `timeout` being set higher than 180s. For large static files, the preboot `python manage.py /app/tmp/collectstatic.py --noinput` can take more than 3 min, causing a failure in the TF. It is recommended to disable the env var, and have the actual collectstatic in a `.profile`.
- Environment variables are passed by the `environment_variables = {}` variable. I wish there was a better way, but each system will have their own ENV specific needs.
- This does not account for platform specific needs, as these may vary per team. This would have to be bound like the others.
```tf
resource "cloudfoundry_user_provided_service" "credentials" {
  name             = "key-service"
  space            = data.cloudfoundry_space.app_space.id
  credentials_json = <<JSON
  {
    "SAM_API_KEY": "${var.sam_api_key}",
    "DJANGO_SECRET_LOGIN_KEY": "${var.django_secret_login_key}",
    "LOGIN_CLIENT_ID": "${var.login_client_id}",
    "SECRET_KEY":"${var.login_secret_key}"
  }
  JSON
}
```
- `service_bindings` will be sent in with a `jsonencode([])` string, and then terraform will `jsondecode()` the string into the proper formatting. Due to this, all services that _would be bound_ to the running app should be deployed first, with a `depends_on = []` containing all references to the services, so the app cannot deploy without those services existing first. As such, it would look like so..
```tf  
module "Application" {
  service_bindings = jsonencode([
    { service_instance = "a_service_name_to_bind" },
    { service_instance = "another_service_name_to_bind" },
    # [...]
    { service_instance = "yet_another_service_name_to_bind" }
  ])
  # [...]
  depends_on = [ module.a_service, module.another_service, module.yet_another_service ]
}
```

**Note**
We expect this to fail, as it is missing key items to fully work. (S3 bucket for submission, creds for login, django keys, etc), but this shows the app atleast available via the uri.
image

## 📜 Testing Plan
This was tested using the following:
```tf
data "cloudfoundry_org" "org" {
  name = var.cf_org_name
}

data "cloudfoundry_space" "space" {
  name = var.cf_space_name
  org  = data.cloudfoundry_org.org.id
}

locals {
  public_s3_name = "fac-public-s3"
  db_name        = "fac-db"
  app_name       = "fac-test"
}

module "s3-public" {
  source       = "github.com/gsa-tts/terraform-cloudgov//s3?ref=v2.1.0"
  cf_space_id  = data.cloudfoundry_space.space.id
  name         = local.public_s3_name
  s3_plan_name = "basic-public"
  tags         = ["s3"]
}

module "database" {
  source        = "github.com/gsa-tts/terraform-cloudgov//database?ref=v2.1.0"
  cf_space_id   = data.cloudfoundry_space.space.id
  name          = local.db_name
  tags          = ["rds"]
  rds_plan_name = "medium-gp-psql"
  json_params = jsonencode(
    {
      "storage" : 25,
    }
  )
}

module "Application" {
  source               = "../application"
  cf_org_name          = var.cf_org_name
  cf_space_name        = var.cf_space_name
  name                 = local.app_name
  branch_name          = "main"
  github_org_name      = "gsa-tts"
  github_repo_name     = "fac"
  src_code_folder_name = "backend"
  buildpacks           = ["https://github.com/cloudfoundry/apt-buildpack.git", "https://github.com/cloudfoundry/python-buildpack.git"]
  environment_variables = {
    DISABLE_COLLECTSTATIC = 1
    ENV                = "SANDBOX"
    DJANGO_BASE_URL    = "https://${local.app_name}.app.cloud.gov"
    ALLOWED_HOSTS      = "${local.app_name}.app.cloud.gov"
    REQUESTS_CA_BUNDLE = "/etc/ssl/certs/ca-certificates.crt"
  }
  service_bindings = {
      (var.service_1_name) = ""
      (var.service_2_name) = "...JSON string..."
  }
  depends_on = [module.database, module.s3-public]
}

```

- deploy a system to a sandbox space


application/prepare_app.sh
```diff
@@ -0,0 +1,31 @@
...
+curl -s -L "https://github.com/${ORG}/${REPO}/archive/${GITREF}.zip" --output local.zip
+branch=$(echo "$GITREF" | cut -f3 -d"/")
+
+zip_folder=$(unzip -l local.zip | awk '/\/$/ {print $4}' | awk -F'/' '{print $1}' | cut -f1 -d"-" | sort -u)
+# Zip up just the $REPO-$branch/ subdirectory for pushing
+# Before zip stage, run [ npm ci --production | npm run build ] in /backend/ to get the compiled assets for the site in /static/compiled/
+unzip -q -u local.zip \*"$zip_folder-$branch/$SRC_FOLDER/*"\*
+cd "${tmpdir}/$zip_folder-$branch/$SRC_FOLDER/" &&
+npm ci --production --silent &&
+npm run build > '/dev/null' 2>&1 &&
```
asteel-gsa, line 25: Not exactly sure this is _fully_ necessary for every team, but it might not hurt?

application/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  memory     = var.app_memory
+  disk_quota = var.disk_space
+  instances  = var.instances
+  strategy   = "rolling"
+
+  routes = [{
+    route = local.app_route
+  }]
+
+  service_bindings = [
```
mogul, line 39: Rather than assuming these existence of these two hard-coded services names, the module var should take an group of service names, and expand that out to as many lines as are necessary here. 

The structure passed in via the module var should allow specifying binding params too; let's get that right now rather than having to handle change-management of that later.

application/variables.tf
```diff
@@ -0,0 +1,74 @@
...
+  default = "2048M"
+}
+
+variable "instances" {
+  type        = number
+  description = "The number of instances for the application"
+  default     = 1
+}
+
+variable "public_s3_name" {
```
mogul, line 66: This should be something like 
```suggestion
variable "service_instance_names" {
```

asteel-gsa: I was toying with this yesterday, but couldn't get it working, think I was just able to..
```tf
# from the tf plan
      + service_bindings            = [
          + {
              + service_instance = "fac-db"
            },
          + {
              + service_instance = "fac-public-s3"
            },
        ]
```
And I managed to do that with [this commit](https://github.com/GSA-TTS/terraform-cloudgov/pull/69/commits/18389d206e597af5767bd8b3db6b9086c16c158e)

So the calling module takes a jsonencoded string, and then decodes it for the `service_bindings` back into an object that it can read, which, now makes it work, and the consumer of the module can make that json as big as they want, and it will set the object correctly
image
Passed in like so:
```tf
  service_bindings = jsonencode([
    {
      service_instance = "${local.db_name}"
    },
    {
      service_instance = "${local.public_s3_name}"
    }
  ])
```

asteel-gsa:
Not sure why the tests are failing, passing on my local:

```tf
[08:08 AM][asteel][application]$ terraform test
tests/creation.tftest.hcl... in progress
  run "application_tests"... pass
  run "src_tests"... pass
tests/creation.tftest.hcl... tearing down
tests/creation.tftest.hcl... pass

Success! 2 passed, 0 failed.
```

application/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+resource "cloudfoundry_app" "application" {
+  name       = var.name
+  space_name = var.cf_space_name
+  org_name   = var.cf_org_name
+
+  path             = "${path.module}/${data.external.app_zip.result.path}"
+  source_code_hash = filesha256("${path.module}/${data.external.app_zip.result.path}")
```
asteel-gsa, line 27: Getting a failure on `filesha256()` when run via github actions, any ideas on this @rahearn ? Is this a case where we want to use the actual provider like the `drupal` module? I would prefer not to if possible, as it is a 3ish min process to send it..

I thought it was an `npm` issue on the runner, so added that, I don't suppose it could be an issue with `awk` or `sort` on [L18?](https://github.com/GSA-TTS/terraform-cloudgov/pull/69/files#diff-34e8b13049a3dc2d7f3829beac2e1da5cf052c51e82ca57318e8f8108b215c80R18)

rahearn: My guess is that there's a difference in output directory between local use and the runner. Can you try updating the prepare script to return the absolute path to the zip file instead of just the filename?

asteel-gsa: Oooo, it worked.. I guess its an npm thing..
@rahearn

application/prepare_app.sh
asteel-gsa: Well, looks like this is no longer going to work, breaks the tests on github.

Team will have to be responsible for getting their staticfiles published as part of a `.profile`

application/main.tf
```diff
@@ -37,7 +36,12 @@ resource "cloudfoundry_app" "application" {
     route = local.app_route
   }]
 
-  service_bindings = jsondecode(var.service_bindings)
-  environment      = merge({}, var.environment_variables)
+  service_bindings = [
+    for service_name, params in var.service_bindings : {
```
mogul, line 40: I don't want to admit how long it took me to get this right, even with an AI assist.

application/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+
+variable "github_repo_name" {
+  description = "The name of the github repo (ex. fac, terraform-cloudgov, etc)"
+  type        = string
+}
+
+variable "src_code_folder_name" {
+  description = "The name of the folder that contains your src code. Generally the folder that would contain your Procfile. This will be used as the apps /app/ dir."
+  type        = string
+}
```
asteel-gsa, line 69: I need to push a check for this to account for systems that use the root of the repo for their system.

application/main.tf
asteel-gsa: @mogul I was only able to get this to work with the following input, is this the type of input you were expecting?

```tf
  service_bindings = {
    "${local.db_name}" = "",
    "${local.public_s3_name}" = "",
  }
```

mogul: Hard to tell what you're referring to... Here's what I notice about that example:

- For the left side of the assignment, it used to be necessary to write `"${varname}"` to have non-static key names, but my understanding is that since a big was fixed you should now only need to do `(varname)`.
- For the right side of the assignment, the string is now expected to just be a JSON string. An empty string `""` is not a valid JSON string, so I added a trigraph that substitutes `"{}"` for that case. 
- It should not be necessary for the module to `jsonencode()` the right side of the assignment to create the actual `params` value anymore; it's always expected to be a JSON string from the module's point of view. I'm mobile so I can't see the code easily but the `jsonencode()` appearing in the plan tells me that I missed removing that somewhere.
- Finally, it shouldn't be necessary to specify an assignment in the map at all if there are no parameters. So this should be working even if you remove `= ""`.

mogul: I guess more directly: That input should work with what I wrote, but it's also unnecessarily complicated.

mogul: This is the equivalent input form that I expect will be most common: 

```hcl
service_bindings = {
    (local.db_name),
    (local.public_s3_name)
  }
```

mogul: > - It should not be necessary for the module to `jsonencode()` the right side of the assignment to create the actual `params` value anymore; it's always expected to be a JSON string from the module's point of view. I'm mobile so I can't see the code easily but the `jsonencode()` appearing in the plan tells me that I missed removing that somewhere.

Just checked and confirmed there is no `jsonencode()` at line 42, so that's being added into the plan by the provider. It's no big deal, as it should be a no-op, but it may be a little distracting. It does imply that the right side assignment in the map _could_ be an HCL object instead of a JSON string, which is friendly, but I'd rather focus camera on the fact that it's a JSON string that will be passed to the bind operation, just as it is when they manually call `cf bind-service` at the CLI... The same string should work in both places.

mogul: Yeah, that's actually right, you must specify a value (it won't default to ""). I think this is a crime, but I have pushed another commit that properly documents this.

asteel-gsa:
> What's left on this one?

As far as I am aware, just @rahearn double checking and deploying to a sandbox space to validate!

application/variables.tf
```diff
@@ -0,0 +1,91 @@
...
+# }
+variable "service_bindings" {
+  description = "A map of service instance name to JSON parameter string."
+  type        = map(string)
+  default     = {}
+}
+
+variable "buildpacks" {
+  description = "A list of buildpacks to add to the app resource."
+  type        = set(string)
```
rahearn, line 41: [issue] `buildpacks` needs to be a `list` because `set` does not preserve order, [while the order of buildpacks is very important](https://github.com/cloudfoundry/terraform-provider-cloudfoundry/issues/84)

application/variables.tf
```diff
@@ -0,0 +1,91 @@
...
+  default     = []
+}
+
+variable "name" {
+  description = "The name of the application to deploy"
+  type        = string
+}
+
+variable "branch_name" {
```
rahearn, line 50: [question] can this just be `ref` instead to potentially deploy tags and or shas?

application/variables.tf
```diff
@@ -0,0 +1,91 @@
...
+  # Examples:
+  # "" -> Project to deploy is in the root of the repo
+  # "backend" -> Project to deploy is in the backend/ directory
+  # "backend/app" -> Project to deploy is in the backend/app directory
+}
+
+variable "app_memory" {
+  type        = string
+  description = "Memory to allocate to app, including unit"
+  default     = "2048M"
```
rahearn, line 78: [suggestion] this seems like an awfully high default (and will prevent this module from working in sandbox spaces out of the box)

application/tests/creation.tftest.hcl
```diff
@@ -0,0 +1,66 @@
...
+  assert {
+    condition     = output.app_id == cloudfoundry_app.application.id
+    error_message = "Output id must match the app id"
+  }
+  assert {
+    condition     = "${var.name}.app.cloud.gov" == output.endpoint
+    error_message = "Endpoint output must match the app route endpoint"
+  }
+  assert {
+    condition     = lookup(cloudfoundry_app.application.environment, "REQUESTS_CA_BUNDLE", "/etc/ssl/certs/ca-certificates.crt") != null
```
rahearn, line 43: [issue] I think this test will always pass? Won't `lookup` return the default value here if the env variable actually is null?

application/main.tf
```diff
@@ -0,0 +1,47 @@
...
+    route = local.app_route
+  }]
+
+  service_bindings = [
+    for service_name, params in var.service_bindings : {
+      service_instance = service_name
+      params           = (params == "" ? "{}" : params) # Empty string -> Minimal JSON
+    }
+  ]
+  environment = merge({}, var.environment_variables)
```
rahearn, line 45: [suggestion] should we add some common env vars needed here, like the SSL cert settings? If not, then we should simplify down to `environment = var.environment_variables`

README.md
```diff
@@ -240,8 +240,71 @@ module "Database" {
...
+      module "Application" {
+        # [...]
+
+        service_bindings = {
+          "a_service","       # hardcoded service name
+        }
+
+        # [...]
+        depends_on = [
+          module.a_service    # modulce creates service "a_service" first
```
rahearn, line 277:
```suggestion
          module.a_service    # module creates service "a_service" first
```

rahearn (changes requested):
The only thing absolutely blocking is the `buildpacks` type.

application/providers.tf
```diff
@@ -0,0 +1,13 @@
...
+  required_version = "~> 1.0"
+  required_providers {
+    cloudfoundry = {
+      source  = "cloudfoundry/cloudfoundry"
+      version = ">=1.3.0"
+    }
+    docker = {
+      source  = "kreuzwerker/docker"
+      version = "~>3.0.2"
+    }
```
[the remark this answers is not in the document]
asteel-gsa: good catch! Im not sure why I copied that in

application/variables.tf
asteel-gsa: Good call!

application/variables.tf
asteel-gsa: So, just tested the curl, and it does seem to have the same format of `REPO-branch`, where the zip file is `REPO-sha`, so yea, I can change that back.

application/variables.tf
asteel-gsa: This was the default for FAC, as we use 1.5G mem at minimum. Maybe we pull out the default and have it passed in?

application/tests/creation.tftest.hcl
asteel-gsa: Yea, I guess if it is a default, and will always pass, no need to test it

application/main.tf
asteel-gsa: Nice call, adding cert bundle to the defaults, with the rest being app specific, so needing to be passed in

README.md
asteel-gsa: fixed

rahearn (approved):
Looks great!

application/variables.tf
mogul: oh wow, good catch here!

application/variables.tf
mogul: Most apps run comfortably within 512M, and you're not losing much by keeping a default at a level that can fit in the 1GB sandbox.

mogul:
Hot damn

rahearn:
> eg in the egress proxy

good idea - I'm gonna take a crack at egress proxy right now

rahearn:
> good idea - I'm gonna take a crack at egress proxy right now

Not compatible out of the box, with at least two issues. One is small and I'll have a PR up to address it. The other is also small but curious as to folk's thoughts on the best path forward:

Issue: the egress proxy's `cloudfoundry_app` specifies the `command` attribute to tell it how to start the caddy server
Potential solutions:

1. Add an optional `command` attribute to the application module, simple, though I'd want to triple check what happens with a null or blank string value and how it interacts with buildpacks that use Procfiles
2. Add a `Procfile` to the cg-egress-proxy repo and don't worry about the command in terraform (at least for now)
3. Really expand this module with support for processes (probably need to do this eventually anyway)

@mogul and @asteel-gsa thoughts on the best route forward?

mogul:
I'd just do `command` for now and wait until we have a real driver for multiple processes to be exposed. Besides, people can always bypass the module and do that by creating a resource with the provider directly; modules should just make the common case the easy case.

Issue #40, opened 2024-05-14
Incorporate FAC Logshipper into Shared Modules


asteel-gsa:
As discussed with Ryan we have a module that we built called the [logshipper](https://github.com/GSA-TTS/FAC/tree/main/terraform/shared/modules/cg-logshipper). This is the terraform module that references the [cg-logshipper](https://github.com/gsa-TTS/cg-logshipper) in order to get the cgov system logs sent to new relic.

The idea of this logshipper application is as follows:
- Create a logshipper application
- Bind any necessary services to the logshipper
- Generate a `logdrain` sysurl
- Bind the drain to the core running application
- Stream the `firehose` to newrelic logs

### Considerations:
- a `logshipper-network-policy` will need to be created from `logshipper` to `https_proxy`
- [`fluentbit.conf`](https://github.com/GSA-TTS/FAC/blob/main/terraform/shared/modules/cg-logshipper/fluentbit_config/fluentbit.conf) is custom for our application, though it does not deviate heavily from the upstream version.
- We give the logshipper a dedicated `s3` so that the logs can be stored in a dedicated bucket
- New Relic credentials are stored as a `cloudfoundry_user_provided_service` and will need to be configured.
- The logshipper uses its own set of `HTTP_USER` and `HTTP_PASS` credentials

Ticket is subject to change as this is a temporary placeholder to give a brief overview.

PR #70, opened 2025-03-18
Add logshipper module

asteel-gsa:
# 🎫 Addresses issue: #40 

## 🛠 Summary of changes
1. Adds a small `logshipper` application
2. This application is heavily reliant on legacy terraform from the `cloudfoundry-community` provider, where it sets up the following:
```
- An s3 service key so that it may be given to GSA SoC for ATO requirements.
- A credential service for the logshipper itself that contains an `HTTP_USER` and `HTTP_PASS`.
- A credential service for new relic, seperate from an existing one that contains a `NEW_RELIC_LICENSE_KEY` and `NEW_RELIC_LOGS_ENDPOINT`, which defaults to the FedRAMP'd endpoint.
- A service called `logdrain` that is used to bind to an application that contains a `syslog_drain_url`
```
## 📜 Testing Plan

```tf
module "logshipper" {
  source                = "../logshipper"
  name                  = var.name
  cf_org_name           = var.cf_org_name
  cf_space = {
    id   = data.cloudfoundry_space.space.id
    name = var.cf_space_name
  }
  https_proxy_url       = var.https_proxy_url
  new_relic_license_key = var.new_relic_license_key
}
```

> [!NOTE]
> An egress proxy, a network policy and a valid new relic license key are **NOT** required to test this module, or even have it operational. They should be added by the consuming team as a means to have the logdrain operating as it was intended, but is not necessary if you are willing to overlook the STDERR's it will generate. All the module needs to function is a binding between the `logdrain` service and a running application. For reasons that still remain unknown to me, a restart of the logshipper may be necessary if the app was created before the logshipper/drain, as doing so will *nudge* the logshipper and make the app realize that the drain is streaming to it 

- Deploy the logshipper module to a sandbox space
- Bind the `logdrain` to a dummy application
- Look inside the `${var.name}-logs-storage` s3 bucket and validate that syslogs and stdout are being drained properly. 

## 👀 Screenshots and Evidence
image


asteel-gsa:
Hmm, not sure why the cg_space tests are failing 😕

asteel-gsa:
I want to write a `null_resource` module for this as well, which does all the things the legacy provider does, just so we can rip it out and not be reliant on the community provider. Ill take care of that tomorrow

rahearn:
> Hmm, not sure why the cg_space tests are failing 😕

Looks like a previous cg_space run didn't finish its cleanup, so the newer test runs couldn't create a space with the same name. I manually removed the space and reran the test and it's fine now.

asteel-gsa:
@rahearn So, I made a `null_resource` to handle all the services and bindings, instead of using the community provider, just so that we don't have terraform that uses it, in the event these resources don't get ported over before the depreciation. We would ideally like to have all services in there, but I defer to you on what you deem more appropriate. Use the `null_resource` or use the community `resources`. Which do you feel more comfortable incorporating? [logshipper-meta.sh](https://github.com/GSA-TTS/terraform-cloudgov/blob/f03f3d0d80d168868d7acb5638ce788da7a4ed64/logshipper/logshipper-meta.sh). The community resources are commented out, if you like the `null_resource` method, I will remove the comments in a new commit.

Along with that, I would really like to offload the s3 from the module, pass that and the binding on to the consuming team, and go back to the `mock_provider` as the tests get a lot of errors with the actual provider creating/deleting the s3. Do you agree with that, or would you prefer to keep the s3 and its binding in this module, minimizing what a team has to do?

mogul:
> Along with that, I would really like to offload the s3 from the module, pass that and the binding on to the consuming team, and go back to the `mock_provider` as the tests get a lot of errors with the actual provider creating/deleting the s3. Do you agree with that, or would you prefer to keep the s3 and its binding in this module, minimizing what a team has to do?

:foop:
... _always opt for dependency-injection_ ...
:poof:

asteel-gsa:
Alright, after going back and forth with this, I think I have it at a stable state. 

I pulled the s3 out of the module and is now a team creation item. AFAIK, it doesn't _need_ an s3, we had it there to provide the syslogs to GSA SoC as part of an ATO requirement. Readme was updated to include that, and basically, the team could create the s3 and then bind it with a null resource, just as we do for the other things.

As much as I wish the provider had a way to bind services, if we want to keep this prod ready, with 0 use of the community provider, we have to use a null_resource. I attempted to build logic around the services, but the logic checks failed with the outputs of `cf service $service_name`. I tried multiple ways to get the error code, parse the output of the return, all of which failed. What happens is it will attempt to create all the time. Not ideal, but if it fails, that apparently does **not** impact the `set -e`, so I am unsure if that return gives a non-zero return code. This is basically done to ensure no services get pulled out from under the terraform.

rahearn:
> Use the `null_resource` or use the community `resources`. Which do you feel more comfortable incorporating?

Was there something that the official provider doesn't support? It looks like you're just using `cups` which can be created already in the official provider. See [this example](https://github.com/GSA-TTS/gitlab-runner-cloudgov/blob/ee1f605b25bc5832f7d48302faf2fa75aa8da64b/main.tf#L187-L200)

asteel-gsa:
> > Use the `null_resource` or use the community `resources`. Which do you feel more comfortable incorporating?
> 
> Was there something that the official provider doesn't support? It looks like you're just using `cups` which can be created already in the official provider. See [this example](https://github.com/GSA-TTS/gitlab-runner-cloudgov/blob/ee1f605b25bc5832f7d48302faf2fa75aa8da64b/main.tf#L187-L200) for creating the service, and [these lines](https://github.com/GSA-TTS/gitlab-runner-cloudgov/blob/ee1f605b25bc5832f7d48302faf2fa75aa8da64b/main.tf#L128-L135) for binding it to the app, and ensuring the app waits for it to be created first

That was **exactly** what I was looking for! But I didn't see it in the docs, as my brain was so used to a seeing it as "user_provided_service", rather than passing ` type = "user-provided"`. Yup! Just need `cups` so I will make that change tomorrow and remove the null resource.

logshipper/logshipper-meta.sh
mogul, line 1: Not reviewing this as I assume Ryan's comment about the user-provided service type has obsoleted it.

logshipper/main.tf
```diff
@@ -0,0 +1,125 @@
+locals {
+  username     = random_uuid.username.result
+  password     = random_password.password.result
+  syslog_drain = "https://${local.username}:${local.password}@${cloudfoundry_route.logshipper_route.host}.app.cloud.gov/?drain-type=all"
```
mogul, line 4: Since I would expect the casual auditor to wonder why the logs endpoint is not private, I'd add a comment hear clarifying that the log-drain must be on the public rather than private app domain to be registered with cloud.gov

logshipper/main.tf
```diff
@@ -0,0 +1,125 @@
...
+  domain       = cloudfoundry_route.logshipper_route.domain
+  app_id       = cloudfoundry_app.logshipper.id
+  #logdrain_id  = cloudfoundry_user_provided_service.logdrain_service.id
```
mogul, line 7:
```suggestion
```

logshipper/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+output "syslog_drain_url" {
+  value     = local.syslog_drain
+  sensitive = true
+}
+
+output "domain" {
+  value = local.domain
+}
+
+# output "logdrain_service_id" {
```
mogul, line 19: Maybe output `logdrain_service_name`, the better for callers to use for to create an implicit dependency.

logshipper/main.tf
mogul, line 1: Not reviewing this in detail since I expect it change a bunch once there's HCL that does the UPS provisioning via the new provider.

logshipper/prepare-logshipper.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Grab a copy of the zip file for the specified ref
+# https://github.com/GSA-TTS/cg-logshipper/archive/refs/heads/main.zip
+curl -s -L https://github.com/GSA-TTS/cg-logshipper/archive/"${GITREF}".zip --output "${tmpdir}/logshipper-dist.zip"
+branch=$(echo "$GITREF" | cut -f3 -d"/")
+
+# Remove the leading directory; the .zip needs to have the files at the top
+cd "${tmpdir}"
+unzip -o "${tmpdir}/logshipper-dist.zip" > /dev/null
+cat "${popdir}/fluentbit_config/fluentbit.conf" > "${tmpdir}/cg-logshipper-main/fluentbit.conf"
```
mogul, line 22: The only part of this file that's specific to the log-shipping context is copying the conf file into the expected place in the .zip file. That seems like something which users of the `app` module might want to do too.

Can we add that feature to the new `app` module so we can use that module to deploy the logshipper instead of using this custom code?

README.md
```diff
@@ -302,6 +302,68 @@ module "Application" {
 }
 ```
 
+## Logshipper
+Creates a logshipper application with all necessary components. It **does not** create a network policy between itsself and your proxy, and does not bind itsself to your desired application. Since there is no terraform to do that binding at this time, you must either `cf bind-service <my-app> logdrain`, supply `logdrain` to your `manifest.yml` or supply `logdrain` to your `service_bindings = {}` for your application.
```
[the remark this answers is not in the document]
asteel-gsa: I just pushed a commit, but I will make it a variable with a default value of "logdrain"

README.md
```diff
@@ -302,6 +302,68 @@ module "Application" {
...
+
+- Logshipper Credentials, defined as `logshipper-creds` is populated as a random user/pass generated via terraform that signifies the `${HTTP_USER}` AND `${HTTP_PASS}` credential set for the `syslog_drain`
+- New Relic Credentials, defined as `logshipper-newrelic-creds` contains your New Relic License Key, and the FedRAMP endpoint for New Relic log API.
+- The s3 bucket, defined as `${var.name}-logs-storage` with a tag of `logshipper-s3` is a dedicated s3 bucket that's sole purpose is the ingest and storage of all `fluentbit` logs being sent.
```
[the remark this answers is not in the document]
asteel-gsa: Fixed, modified to `defined as <my-s3-name>`

README.md
```diff
@@ -302,6 +302,68 @@ module "Application" {
...
+  name        = var.name
+  cf_org_name = var.cf_org_name
+  cf_space = {
+    id   = data.cloudfoundry_space.space.id
+    name = var.cf_space_name
+  }
+  logshipper_s3_name    = local.logshipper_storage_name
+  https_proxy_url       = var.https_proxy_url
+  new_relic_license_key = var.new_relic_license_key
+  depends_on = [module.logs-storage]
```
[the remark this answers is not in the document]
asteel-gsa: I just pushed a commit that _should_ make this work. Though, I do think we should put the outputs of `bucket_name` in the outputs of the s3 module

```tf
  service_bindings = {
    "${local.logshipper_storage_name}" = ""
  }
  depends_on = [module.logs-storag
```

logshipper/fluentbit_config/fluentbit.conf
```diff
@@ -0,0 +1,112 @@
...
+    preserve_data_ordering On
+
+# Ship to New Relic:
+[OUTPUT]
+    Name newrelic
+    Match *
+    licenseKey ${NEW_RELIC_LICENSE_KEY}
+    endpoint ${NEW_RELIC_LOGS_ENDPOINT}
+
+# Uncomment to see the parsed messages in the logshipper's logs:
```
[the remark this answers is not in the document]
asteel-gsa: Looks like ours in the FAC differs from Amy's. This is the FAC's config, im going to ditch the changing of the conf and just use Amy's without modification.

logshipper/main.tf
asteel-gsa: Added a comment with refs to docs

logshipper/main.tf
asteel-gsa: Deleted as stray code!

logshipper/outputs.tf
asteel-gsa: Changed output of id to output of drain name

asteel-gsa:
So, I modified the module, since now I can use the `cups` resource 😄, and did a cheeky thing for binding the services. I really like this new change that we did in the app module..

```tf
locals {
  services = merge({
    "${local.logshipper_creds}"    = ""
    "${local.newrelic_creds_name}" = ""
  }, var.service_bindings)
}

resource "cloudfoundry_app" "logshipper" {
  service_bindings = [
    for service_name, params in local.services : {
      service_instance = service_name
      params           = (params == "" ? "{}" : params) # Empty string -> Minimal JSON
    }
  ]
}
```

TF Plan
```
      + service_bindings            = [
          + {
              + params           = jsonencode({})
              + service_instance = "logshipper-creds"
            },
          + {
              + params           = jsonencode({})
              + service_instance = "logshipper-logs"
            },
          + {
              + params           = jsonencode({})
              + service_instance = "logshipper-newrelic-creds"
            },
        ]
```

asteel-gsa:
oop, looks like we have some stray services in the `ci-tests` space 🤣 
```
Error: API Error in creating user-provided service instance

  with cloudfoundry_service_instance.logshipper_creds,
  on main.tf line 86, in resource "cloudfoundry_service_instance" "logshipper_creds":
  86: resource "cloudfoundry_service_instance" "logshipper_creds" {

Unable to create service instance logshipper-creds: executing POST request
for /v3/service_instances failed: cfclient error
(CF-UnprocessableEntity|10008): The service instance name is taken:
logshipper-creds.

Error: API Error in creating user-provided service instance

  with cloudfoundry_service_instance.logdrain,
  on main.tf line 112, in resource "cloudfoundry_service_instance" "logdrain":
 112: resource "cloudfoundry_service_instance" "logdrain" {

Unable to create service instance logdrain: executing POST request for
/v3/service_instances failed: cfclient error (CF-UnprocessableEntity|10008):
The service instance name is taken: logdrain.
```

README.md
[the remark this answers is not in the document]
asteel-gsa: https://github.com/GSA-TTS/terraform-cloudgov/pull/71

So if we want to make this change, we can.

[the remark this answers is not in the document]
mogul: I merged that. Let's document by referring to it!

logshipper/prepare-logshipper.sh
asteel-gsa: Not sure we can do this as is, since logshipper does rely on a sidecar, and I am not exactly sure how an app would respond with an empty set for the sidecar if we default it to fully empty, as it does require some input [validator](https://github.com/cloudfoundry/terraform-provider-cloudfoundry/blob/18aa58334865d548d7fba1dac225cf369e19d402/internal/provider/resource_app.go#L239C1-L241C7)

README.md
```diff
@@ -302,6 +302,68 @@ module "Application" {
...
+```tf
+module "logshipper" {
+  source      = "github.com/gsa-tts/terraform-cloudgov//logshipper?ref=v2.5.0"
+  name        = var.name
+  cf_org_name = var.cf_org_name
+  cf_space = {
+    id   = data.cloudfoundry_space.space.id
+    name = var.cf_space_name
+  }
+  logshipper_s3_name    = local.logshipper_storage_name
```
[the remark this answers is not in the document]
asteel-gsa: This should have been removed

README.md
```diff
@@ -302,6 +302,60 @@ module "Application" {
...
+  cf_space = {
+    id   = data.cloudfoundry_space.space.id
+    name = var.cf_space_name
+  }
+  https_proxy_url       = var.https_proxy_url
+  new_relic_license_key = var.new_relic_license_key
+  service_bindings = {
+    "${module.logs-storage.bucket_name}" = ""
+  }
+  depends_on = [module.logs-storage]
```
mogul, line 341:
```suggestion
```

mogul (dismissed):

logshipper/providers.tf
```diff
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     cloudfoundry = {
       source  = "cloudfoundry/cloudfoundry"
-      version = "1.3.0"
+      version = "1.4.0"
```
rahearn, line 6: [suggestion] specify as `">=1.4.0"` so we don't run into version issues with multiple modules

asteel-gsa: Just pushed that update!

asteel-gsa:
Should be all set for review. Deploys cleanly with the new route module @rahearn

app_route/outputs.tf
```diff
@@ -5,3 +5,7 @@ output "endpoint" {
 output "route_id" {
   value = cloudfoundry_route.app_route.id
 }
+
+output "host" {
+  value = cloudfoundry_route.app_route.host
+}
```
rahearn, line 11: [suggestion] I'm not opposed to including this, but I don't think it's necessary to add

logshipper/main.tf
```diff
@@ -0,0 +1,116 @@
+locals {
+  username = random_uuid.username.result
+  password = random_password.password.result
+  # The syslog_drain must be registered on the public domain for cloudfoundry.
+  # Cloud Foundry uses the syslog URL to route messages to the service.
+  # The syslog URL has a scheme of syslog, syslog-tls, or https, and can include a port number.
+  # More information here:
+  # https://docs.cloudfoundry.org/devguide/services/log-management.html
+  # https://docs.cloudfoundry.org/devguide/services/user-provided.html#syslog
+  syslog_drain = "https://${local.username}:${local.password}@${module.route.host}.app.cloud.gov/?drain-type=all"
```
rahearn, line 10: [issue] if `var.domain` is passed into the module, then this url will be wrong.

```suggestion
  syslog_drain = "https://${local.username}:${local.password}@${module.route.endpoint}/?drain-type=all"
```

logshipper/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  app_id       = cloudfoundry_app.logshipper.id
+  route        = "${var.cf_space.name}-${var.name}.app.cloud.gov"
```
rahearn, line 12: [suggestion] looks like this isn't used anymore

```suggestion
```

logshipper/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  services = merge({
+    "${local.logshipper_creds_name}" = ""
+    "${local.newrelic_creds_name}"   = ""
+  }, var.service_bindings)
+}
+
+data "cloudfoundry_domain" "public" {
+  name = "app.cloud.gov"
+}
+
```
rahearn, line 26: [suggestion] looks like this isn't used anymore

```suggestion
```

logshipper/main.tf
```diff
@@ -0,0 +1,116 @@
...
+  syslog_drain_url = local.syslog_drain
+}
+
+module "route" {
+  source = "../app_route"
+
+  cf_org_name   = var.cf_org_name
+  cf_space_name = var.cf_space.name
+  domain        = var.domain
+  hostname      = coalesce(var.hostname, var.name)
```
rahearn, line 114: [suggestion] I think we should prevent "default" use from having such easy route collisions. I'm thinking something like: `hostname = coalesce(var.hostname, random_pet.host.id)` using [random_pet provider](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/pet)

logshipper/tests/creation.tftest.hcl
```diff
@@ -0,0 +1,46 @@
+provider "cloudfoundry" {
+  api_url = "https://api.fr.cloud.gov"
+  # cf_user and cf_password are passed in via CF_USER and CF_PASSWORD env vars
+}
```
rahearn, line 4: [suggestion] we can use cf-cli auth in tests too if we remove all config set here.

```suggestion
provider "cloudfoundry" {}
```

rahearn (changes requested):

logshipper/tests/creation.tftest.hcl
asteel-gsa: Fixed

logshipper/main.tf
asteel-gsa: Added a random_pet route of `logshipper-{pet}`

logshipper/main.tf
asteel-gsa: Removed

logshipper/main.tf
asteel-gsa: Removed

rahearn (dismissed):
:shipit:

asteel-gsa:
Last one, _i think_

PR #72, opened 2025-03-24
Use application module within egress proxy module

rahearn:
# 🎫 Addresses issue: no issue

## 🛠 Summary of changes

* update egress-proxy module to use application module
* tweaks to application module from ☝🏻 
    * allow passing full app route to support `apps.internal` routing
    * allow passing app command


rahearn:
Closing as egress-proxy is too much of a special case to use the app module right now (app module doesn't yet support an app that does not use any service bindings)

PR #73, opened 2025-03-24
App module improvements

rahearn:
# 🎫 Addresses issue: no issue

## 🛠 Summary of changes

* remove default values for variables that are invalid if actually used
* allow specifying the apps route, or fallback to `${var.name}.app.cloud.gov`

### DIFF
--- a/application/variables.tf
+++ b/application/variables.tf
@@ -46,7 +46,7 @@ variable "name" {
 }
 
 variable "route" {
-  description = "The full app route. Defaults to {var.name}.app.cloud.gov if this is omitted"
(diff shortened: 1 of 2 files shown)


application/variables.tf
```diff
@@ -33,20 +33,24 @@ variable "environment_variables" {
...
   type        = list(string)
-  default     = []
 }
 
 variable "name" {
   description = "The name of the application to deploy"
   type        = string
 }
 
+variable "route" {
```
asteel-gsa, line 48: [Non-Blocking] 
We might want to drop an example, just in case. Something like:
```tf
# Should look like my-app-dev.apps.internal or my-app-dev.app.cloud.gov
```

asteel-gsa (approved):
Nice catch! Thank you for this.

PR #76, opened 2025-03-27
Use proper route resources

rahearn:
# 🎫 Addresses issue: no issue

## 🛠 Summary of changes

* updates to provider 1.4 which has a fix for `cloudfoundry_route` resources wanting to be recreated on every apply.
* Uses real resources so that deletion does not leave abandoned routes (which happened with just letting `cloudfoundry_app` implicitly create the route)
* Created an `app_route` module, but haven't documented it yet since I intended it to be internally used. Feel free to tell me that this was the wrong call.
* Updated testing setup to be able to re-use cf-cli authentication and no longer require creating service accounts locally


app_route/main.tf
```diff
@@ -0,0 +1,25 @@
...
+}
+
+data "cloudfoundry_domain" "domain" {
+  name = var.domain
+}
+
+locals {
+  destinations = (length(var.app_ids) == 0 ? null : [
+    for dest in var.app_ids : { app_id = dest }
+  ])
```
rahearn, line 16: this is actually probably the solution for using the application module within egress proxy as well.

asteel-gsa:
Do we not need to bring `domain` up to `>=1.4.0` as well?

rahearn:
> Do we not need to bring `domain` up to `>=1.4.0` as well?

oh, yes - and also we can make some significant improvements to how that one works. Missed it because I was just looking at the cloudfoundry_app resources.

domain/variables.tf
```diff
@@ -11,23 +11,29 @@ variable "cf_space" {
   description = "cloud.gov space resource"
 }
 
-variable "app_names" {
+variable "app_ids" {
```
asteel-gsa, line 14: Suggestion:
We should probably reference this in the `readme` as well with something like 
```tf
data.cloudfoundry_app.app.id
# or
module.app.app_id
```

application/main.tf
```diff
@@ -1,6 +1,5 @@
 locals {
-  app_route = coalesce(var.route, "${var.name}.app.cloud.gov")
-  app_id    = cloudfoundry_app.application.id
+  app_id = cloudfoundry_app.application.id
```
asteel-gsa, line 2:
```suggestion
```
I wonder if we should remove this local and just reference `cloudfoundry_app.application.id` where we need it.

application/outputs.tf
```diff
@@ -3,7 +3,7 @@ output "app_id" {
 }
```
asteel-gsa, line 4:
```suggestion
output "app_id" {
  value = cloudfoundry_app.application.id
}
```

application/main.tf
```diff
@@ -42,3 +37,13 @@ resource "cloudfoundry_app" "application" {
...
 }
+
+module "route" {
+  source = "../app_route"
+
+  cf_org_name   = var.cf_org_name
+  cf_space_name = var.cf_space_name
+  domain        = var.domain
+  hostname      = coalesce(var.hostname, var.name)
+  app_ids       = [local.app_id]
```
asteel-gsa, line 48:
```suggestion
  app_ids       = [cloudfoundry_app.application.id]
```

asteel-gsa:
Initial glance looks good so far, im pretty tired so going to logout and I will do some batch testing of the modules tomorrow.

One more question, should we use this opportunity to force everything up to `>=1.4.0` just so we get it all done in one batch?

domain/variables.tf
```diff
@@ -11,23 +11,29 @@ variable "cf_space" {
...
 
 variable "cdn_plan_name" {
   type        = string
   description = "name of the service plan name to create"
 }
 
+variable "create_domain" {
+  type        = bool
+  default     = false
+  description = "Set to true to create the domain resource. Must be called by an OrgManager"
```
mogul, line 34:
```suggestion
  description = "Set to true to create the domain resource. Requires the OrgManager role."
```

spiffworkflow/main.tf
```diff
@@ -176,10 +189,6 @@ resource "cloudfoundry_app" "frontend" {
   instances         = var.frontend_instances
   strategy          = "rolling"
   health_check_type = "port"
-  routes = [{
-    route    = local.frontend_route
-    protocol = "http1"
```
mogul, line 181: @asteel-gsa didn't we determine that one of these routes needed to be http1? Not seeing that option in @rahearn's new module AFAICS.

mogul (dismissed):

spiffworkflow/main.tf
asteel-gsa: @mogul Im not seeing anything wrong with the config, at least, deploying from this branch. It looks like it got it?

image

asteel-gsa (dismissed):
Deployed both the Application and Spiff modules using the new routing, both look good to me!

mogul:
Thanks for checking!

asteel-gsa:
Once Ryan updates the PR and it is merged, I will rebase the `loghsipper` onto main and swap the routing logic with the new module

app_route/main.tf
rahearn: one of the blockers for using the application module by the egress proxy was that the proxy doesn't bind any services, and the `cloudfoundry_app` module rejects `service_bindings = []` as invalid. Using this ternary to actually set `service_bindings = null` will likely get around that

application/main.tf
```diff
@@ -42,3 +37,13 @@ resource "cloudfoundry_app" "application" {
     REQUESTS_CA_BUNDLE = "/etc/ssl/certs/ca-certificates.crt"
   }, var.environment_variables)
 }
+
+module "route" {
```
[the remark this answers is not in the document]
rahearn: There are a lot of advanced use cases that the application module doesn't yet support. I'm going to put "no-route" use in that bucket as well.

mogul (approved):

PR #81, opened 2025-04-23
Add a note about the proxy and new relic in the readme

asteel-gsa:
## No Issue
Updates the readme for the proxy to include findings from when the FAC upgraded to the module. We use a python new relic agent on the instance to send telemetry along with the logshipper, and when we deployed, we noticed that all traffic was successfully going through the proxy, except for the new relic agent. Allowing both 443 and 61443 fixed the issue, presumably due to the `https_uri` and the new relic agent having the `:61443` port for the proxy string. 

Not setting this as a default, just an addition to the readme as some teams may not want 61443 forced by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,9 @@ Creates and configures an instance of cg-egress-proxy to proxy traffic from your
 
 Prerequite: existing public-egress space to deploy the proxy into
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


rahearn (approved):
thanks! I'm glad you chose the README approach for this

PR #85, opened 2025-07-15
Use stable v1.0.0 SpiffWorkflow image tags

mogul:
Now that all of our changes have been pushed upstream, we can pull directly from the images produced by Sartography, rather than those from our fork.


mogul:
OK, docs added! I'd like to get this approved soon so I can get rid of the forked spiff-arena repo.

asteel-gsa (approved):
LGTM

A trivy scanned image will still be accessible via `ghcr.io/gsa-tts/terraform-cloudgov/<image_name>:latest`. Only the base image changes, which we agreed GSA-TTS maintaining an older image is no longer necessary

PR #86, opened 2025-07-15
Spiff without containers

mogul:
Adds an option to deploy the backend using a buildpack. This skips a lot of hassle needed to deploy when you have custom process_models, because they don't have to be in a public repository or in an image. See [this example of what the typical workflow app will look like](https://github.com/GSA-TTS/pic-blm-cxworks/tree/add-workflow-terraform-module/workflow), and [this call to invoke the module](https://github.com/GSA-TTS/pic-blm-cxworks/blob/d7222f20491859e1212c4631b82437d95924dda3/terraform/apps/workflow/main.tf#L17-L32).

Also adds a detailed `spiffworkflow/README.md` file.


mogul:
Ughhhh, I haven't set up commit-signing on my refreshed machine yet. Will handle that in the morning.

mogul:
Signed using cosign, which uses ephemeral certs, which is why it says "Unverified"... a known issue. Will fix in the morning.

rahearn:
> @rahearn can you help me figure out why the test is failing? The output is not particularly helpful.

It's an issue with a mock_provider not returning IDs of the expected format, but it still validates that the incoming IDs are the correct format.

spiffworkflow/README.md
```diff
@@ -0,0 +1,125 @@
...
+- Deploy all components of SpiffWorkflow in a cloud.gov space
+- Automatic configuration of network policies and routes
+- Two deployment options for the backend:
+  - Container-based deployment using a pre-built Docker image
+  - Buildpack-based deployment from source code using local process models
+
+## Usage
+
+**NOTE:**
+Ensure that your space has the `public_networks_egress`security group if you are not using an egress proxy, so that the frontend can reach the backend API endpoint.
```
asteel-gsa:
```suggestion
Ensure that your space has the `public_networks_egress` and `trusted_local_networks_egress` security groups if you are not using an egress proxy, so that the frontend can reach the backend API endpoint.
```

`trusted_local_networks_egress` is necessary for the app to communicate with the DB. `public_networks_egress` doesn't support rds connections

spiffworkflow/main.tf
```diff
@@ -27,111 +31,10 @@ resource "random_password" "password" {
   special = false
 }
 
-resource "random_pet" "route_prefix" {
+resource "random_pet" "prefix" {
```
asteel-gsa, line 34: This is predominantly now to signify a "non-prod" state. While route collisions are a possibility they are not the main focus. We should probably strip this to be a more normalized context later, i.e `"spiffworkflow-${var.cf_space.name}-${local.app_name}` or something similar

asteel-gsa: This is a non blocker for now, something we can deal with later

asteel-gsa (approved):
Discussed on call with Bret. Some small things were added which we agree on for the general structure.

PR #89, opened 2025-07-24
Buildpack-based spiffworkflow connector

mogul:
# 🎫 Addresses issue: https://github.com/GSA-TTS/pic-blm-cxworks/issues/20

## 🛠 Summary of changes

Adds an option to deploy the connector from a local implementation using buildpacks. This is necessary for us to customize the connector without having to go through a bunch of steps to build an image, push it, and then have CF pull it. 

(In future, we'll use the Paketo CNB for this.)

This is working for me locally, though I haven't yet got a working connector for the buildpack to process. I haven't changed the defaults; it will still deploy Sartography's sample connector image unless told otherwise.


mogul:
Tests are failing because the fix we rely on hasn't made it into a new release of terraform-provider-cloudfoundry yet, but [there should be one soon](https://github.com/cloudfoundry/terraform-provider-cloudfoundry/pull/287/#issuecomment-3301898766).

PR #97, opened 2026-01-21
Remove init process

Mwindo:
These changes were tested here: https://github.com/GSA-TTS/pic-blm-cxworks/issues/461

This is a PR just for a more visible paper trail rather than committing directly to our stable branch. Related to https://github.com/GSA-TTS/spiff-arena/pull/16.


Mwindo:
@mogul, I'm going to go ahead and merge (it won't affect our running app since we are pointing to a specific commit hash) so that I can open a PR pic-blm side pointing to the updated `custom-spiffworkflow-connector-stable` commit hash.

PR #98, opened 2026-01-23
Update to remove init process

Mwindo:
[This PR](https://github.com/GSA-TTS/terraform-cloudgov/pull/97) (merged into the stable branch) worked, but there were no guards for multi-instance deployments; in other words, multiple instances would run the bootstrap process model on startup. To handle this:

1. We add the .profile.d script directory back into the `build-backend.sh` in this PR.
2. PIC-side, in a .profile.d script, we have a script to unset the bootstrap environment variable if the `CF_INSTANCE_INDEX` is not 0.

### DIFF
--- a/spiffworkflow/build-backend.sh
+++ b/spiffworkflow/build-backend.sh
@@ -355,6 +355,12 @@ echo "Process models copied to: ${PROCESS_MODELS_DEST}"
 echo "Removing any .bpmn.png files from process models directory..."
 find "${PROCESS_MODELS_DEST}" -type f -name "*.bpmn.png" -delete
 
+# Ensure we run the bootstrap process model, if it exists at all, on only one app instance
+if [[ "${CF_INSTANCE_INDEX:-0}" != "0" ]]; then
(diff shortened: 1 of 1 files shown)


spiffworkflow/build-backend.sh
```diff
@@ -355,6 +355,23 @@ echo "Process models copied to: ${PROCESS_MODELS_DEST}"
 echo "Removing any .bpmn.png files from process models directory..."
 find "${PROCESS_MODELS_DEST}" -type f -name "*.bpmn.png" -delete
 
+# ----------------------------------------------------------------------------
+# Include content of custom scripts directory (eg profile hooks)
+# Note these are copied relative to the root of the application!
+# ----------------------------------------------------------------------------
+if [ -n "${BACKEND_SCRIPTS_PATH}" ] && [ -d "${BACKEND_SCRIPTS_PATH}" ]; then
```
Mwindo, line 362: Very similar to what we had before, just minus the `init-process.py` script since we now rely on the Spiff bootstrap.

PR #100, opened 2026-02-06
Add prevent_destroy lifecycle option to the database resource

Mwindo:
We have accidentally deleted an RDS resource twice and want to avoid that in the future. This seems like the best way to do so.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -64,6 +64,46 @@ Follow the steps in the [cloudfoundry provider migration guide](https://github.c
 1. Run: `terraform import module.database.cloudfoundry_service_instance.rds ID_FROM_STEP_3`
 1. Run: `terraform apply` to fill in new computed attributes
 
+## Database: `prevent_destroy` support
+
+The database module now supports a `prevent_destroy` variable (default: `false`) that prevents accidental destruction of the database service instance.
+
+### Upgrading existing databases (default behavior, `prevent_destroy = false`)
+
+No action required. The module includes a `moved` block that automatically migrates state from `cloudfoundry_service_instance.rds` to `cloudfoundry_service_instance.rds[0]`. Run `terraform plan` and verify you see only a state address change with zero infrastructure modifications, then `terraform apply`.
+
+### Enabling protection on an existing database (`prevent_destroy = true`)
+
+> ⚠️ **WARNING:** You **must** perform the manual state move below **before** running `terraform plan`. If you skip this step, Terraform will plan to **destroy** the existing database and create a new one. The `terraform plan` output will clearly show `1 to destroy, 1 to add` — always review plans carefully.
+
+1. Update your module source ref to the new version.
+2. Set `prevent_destroy = true` in your module call.
+3. **Before running `terraform plan`**, move the state (replace `<NAME>` with your module's name, e.g., `workflow-db`, `database`):
+
+    ```bash
+    terraform state mv \
+      'module.<NAME>.cloudfoundry_service_instance.rds' \
+      'module.<NAME>.cloudfoundry_service_instance.rds_protected[0]'
+    ```
+
+4. Run `terraform plan` and confirm zero infrastructure changes.
+5. Run `terraform apply`.
+
+### Enabling protection on a database that already upgraded with the default
+
+If you previously upgraded with `prevent_destroy = false` (the default) and later want to enable protection:
+
+1. Set `prevent_destroy = true` in your module call.
+2. Move the state:
+
+    ```bash
+    terraform state mv \
+      'module.<NAME>.cloudfoundry_service_instance.rds[0]' \
+      'module.<NAME>.cloudfoundry_service_instance.rds_protected[0]'
+    ```
+
+3. Run `terraform plan`, confirm zero changes, and `terraform apply`.
+
 ## Module Changes
 
 ### Common Changes
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -68,33 +68,21 @@ Follow the steps in the [cloudfoundry provider migration guide](https://github.c
 
 The database module now supports a `prevent_destroy` variable (default: `false`) that prevents accidental destruction of the database service instance.
 
-### Upgrading existing databases (default behavior, `prevent_destroy = false`)
+### Step 1: Upgrade to the new module version
 
-No action required. The module includes a `moved` block that automatically migrates state from `cloudfoundry_service_instance.rds` to `cloudfoundry_service_instance.rds[0]`. Run `terraform plan` and verify you see only a state address change with zero infrastructure modifications, then `terraform apply`.
+This step is the same for everyone. Update your module source ref to the new version and run:
 
-### Enabling protection on an existing database (`prevent_destroy = true`)
+1. `terraform plan` — verify you see only a state address rename (`cloudfoundry_service_instance.rds` → `cloudfoundry_service_instance.rds[0]`) with zero infrastructure changes.
+2. `terraform apply`.
 
-> ⚠️ **WARNING:** You **must** perform the manual state move below **before** running `terraform plan`. If you skip this step, Terraform will plan to **destroy** the existing database and create a new one. The `terraform plan` output will clearly show `1 to destroy, 1 to add` — always review plans carefully.
+### Step 2 (optional): Enable `prevent_destroy`
 
-1. Update your module source ref to the new version.
-2. Set `prevent_destroy = true` in your module call.
-3. **Before running `terraform plan`**, move the state (replace `<NAME>` with your module's name, e.g., `workflow-db`, `database`):
+If you want to protect a database from accidental destruction, perform this step after completing Step 1.
 
-    ```bash
-    terraform state mv \
-      'module.<NAME>.cloudfoundry_service_instance.rds' \
-      'module.<NAME>.cloudfoundry_service_instance.rds_protected[0]'
-    ```
-
-4. Run `terraform plan` and confirm zero infrastructure changes.
-5. Run `terraform apply`.
-
-### Enabling protection on a database that already upgraded with the default
-
-If you previously upgraded with `prevent_destroy = false` (the default) and later want to enable protection:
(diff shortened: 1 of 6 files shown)


rahearn (changes requested):
This will prevent using the module for review apps, so I don't think this is the appropriate place to put this restriction

Mwindo:
> This will prevent using the module for review apps, so I don't think this is the appropriate place to put this restriction

@rahearn That's true, although right now my preference is for more tedium in ephemeral environments (being forced to manually tear things down when needed) instead of increased risk of wiping out data. That said, I'd be delighted if there is a better, standard way to do this. What would your recommendation be? I'd love to be able to do, e.g.,

```
lifecycle {
  prevent_destroy = var.prevent_destroy
}
```

But the value of `prevent_destroy` needs to be a literal. We could maybe run some sort of script in our deployment pipeline that checks the plan and triggers an "Are you _sure_ about this?" step, it just feels a little hacky.

rahearn:
We already have uses of this module that are used in ephemeral per-merge request deployments with automatic tear-down on merge, so this is a non-starter until terraform does support passing a variable to lifecycle (which I agree would be ideal).

I'd suggest investigating either adding a no-op resource in your root module that prevents deploy of everything and/or removing manual access to running terraform against production from any people and rely on CI-only for deployments.

Mwindo:
> and rely on CI-only for deployments

This happened via CI! (We're still investigating why so we can address it.)

> We already have uses of this module that are used in ephemeral per-merge request deployments with automatic tear-down on merge

The original thought was that this is a custom branch not necessarily meant to be merged into main, although there's really nothing preventing that from happening accidentally down the line, which makes this a dangerous approach.

I suppose another option is to add a parallel "protected database" module with `prevent_destroy = true`, although moving from one module to the other would involve some jiggery pokery. I'll talk with @mogul and @btylerburton and see what we can come up with.

mogul:
@rahearn said:

> this is a non-starter until terraform does support passing a variable to lifecycle

We could work around this by having two resources, one with `prevent_destroy` set to true, and one with it set to false. Then we can use `count` and a bool variable to control which one is actually created. What do you think about that approach?

It might cause problems for migration; not sure if we can write a generic import statement. 🤔

mogul:
Oh, [there's a `moved` block now](https://developer.hashicorp.com/terraform/language/block/moved)!

rahearn:
> Oh, [there's a `moved` block now](https://developer.hashicorp.com/terraform/language/block/moved)!

yep - this is the answer for that.

I actually love that option - keep the default behavior as current, and add a `variable "prevent_destroy"` that creates the protected version if it's true

mogul:
I'll update the branch.

rahearn:
Hmm - `moved` block may not entirely work within the module, since it would probably have the wrong address for a user with `prevent_destroy` set to true.

Even if that's the case, I still think I'm in favor of this path, but it would need to be a new major revision with instructions on what a user would need to move in order to upgrade

mogul:
I tried to signpost everywhere appropriate that opting into this protection for an existing database requires diligence so you don't inadvertently approve a replacement, and that you need to do an advance `terraform mv`. There's still no impact to existing databases as part of the upgrade; I see this as a minor .x release because it's adding a feature. What do you think?

database/main.tf
```diff
@@ -8,10 +8,37 @@ data "cloudfoundry_service_plans" "rds" {
...
+  }
+}
+
+# Automatically migrate existing state for the default (prevent_destroy = false)
+# upgrade path. Previously the resource had no count; adding count requires moving
+# from the unindexed address to [0]. Without this block, Terraform would plan to
+# destroy the existing instance and create a new one.
+#
+# If setting prevent_destroy = true on an existing database, you must manually run
+# `terraform state mv` BEFORE planning. See UPGRADING.md for details.
```
rahearn: [question] have you tried a migration from existing db to one with `prevent_destroy = true`? I'm worried that this moved block would cause an error if the `to` address doesn't exist in the result, but if it deals with that cleanly then hooray

mogul: I'll give it a shot as soon as I have 🥄🥄.

mogul: I'll give it a shot as soon as I have 🥄🥄.

mogul: I let Claude Opus 4.6 test this live (under my command-by-command supervision) against cloud.gov via a real resource. It cleverly suggesting the test use an UPSI to avoid waiting for RDS to provision, which was nice because it completely focused on the Terraform behavior. The answer: The `moved` block doesn't error, and you run right into the scenario warned about: The plan will include destroying the original DB and creating a new one. Screenshot attached. 

Image

mogul: I explored a few other options to see if I could make it:

* error out only while in that as-yet-unmoved state?
  * no
* make the planned destroy itself more obvious
  * yes via a warning message, but the message would be tied to `prevent_destroy == true`, so it would pollute every plan forever.

Given the plan output is:

> cloudfoundry_service_instance.rds[0] will be destroyed
> (because index [0] is out of range for count)
> (moved from cloudfoundry_service_instance.rds)
> Plan: 1 to add, 0 to change, 1 to destroy.

I think we just need to trust people to review the plan as they're supposed to, at least just this once while the migration happens. Hopefully it's the only thing they're changing so the plan will be short and the details hard to glide past; here is what that plan would look like. If they're doing this in a terminal there's even that red text to help make it obvious.

Image

OK to merge?

UPGRADING.md
```diff
@@ -64,6 +64,46 @@ Follow the steps in the [cloudfoundry provider migration guide](https://github.c
...
+    ```bash
+    terraform state mv \
+      'module.<NAME>.cloudfoundry_service_instance.rds' \
+      'module.<NAME>.cloudfoundry_service_instance.rds_protected[0]'
+    ```
+
+4. Run `terraform plan` and confirm zero infrastructure changes.
+5. Run `terraform apply`.
+
+### Enabling protection on a database that already upgraded with the default
```
Mwindo: Minor: this part of the change in the README (here to below in the diff) seems unnecessary and therefore slightly confusing. It is a repeat of the above, just in the case you don't need to do `1. Update your module source ref to the new version.`

rahearn: I agree. I think a clearer upgrade path for folks would be:

1. Upgrade to new module version (prevent_destroy=false / default moved block)
2. Update/move from `rds[0]` to `rds_protected[0]`

That way, step 1 is the same for everyone no matter what, and then step 2 is an optional step for those who want `prevent_destroy`, but it can assume the same starting point. And if someone knows terraform well enough to do it in one step, then kudos to them

UPGRADING.md
```diff
@@ -64,6 +64,46 @@ Follow the steps in the [cloudfoundry provider migration guide](https://github.c
...
+
+The database module now supports a `prevent_destroy` variable (default: `false`) that prevents accidental destruction of the database service instance.
+
+### Upgrading existing databases (default behavior, `prevent_destroy = false`)
+
+No action required. The module includes a `moved` block that automatically migrates state from `cloudfoundry_service_instance.rds` to `cloudfoundry_service_instance.rds[0]`. Run `terraform plan` and verify you see only a state address change with zero infrastructure modifications, then `terraform apply`.
+
+### Enabling protection on an existing database (`prevent_destroy = true`)
+
+> ⚠️ **WARNING:** You **must** perform the manual state move below **before** running `terraform plan`. If you skip this step, Terraform will plan to **destroy** the existing database and create a new one. The `terraform plan` output will clearly show `1 to destroy, 1 to add` — always review plans carefully.
```
rahearn: [suggestion] re-word this warning to focus on what they _should_ see.


```suggestion
> ⚠️ **WARNING:** You **must** perform the manual state move below **before** running `terraform plan`. If you skip this step, Terraform will plan to **destroy** the existing database and create a new one. `terraform plan` should clearly include `0 to destroy` if you've done the move correctly — always review plans carefully.
```

rahearn (changes requested):
Code is great, I love this a lot and wonder if we should bring it to some of the other items as well.

Before merge, can we update `UPGRADING` and also change the target to merge into `main` instead of the spiffworkflow branch?

rahearn:
> I see this as a minor .x release because it's adding a feature. What do you think?

yes, with the default moved block I also see this as a minor .x release.

mogul (dismissed):

mogul:
@rahearn any idea what's up with that failing cg_space integration test?

cg_space/tests/creation.tftest.hcl
```diff
@@ -1,5 +1,43 @@
 provider "cloudfoundry" {}
 
+override_data {
```
mogul: > @rahearn any idea what's up with that failing cg_space integration test?

nvm, Claude figured out how to mock out the API calls and the need for access to a particular org.

mogul:
Doh!

> tests/creation.tftest.hcl... in progress
>   run "test_space_creation"... pass
>   run "test_manager_only"... pass
>   run "test_individual_permissions"... fail
>   run "test_deployer_permissions"... skip
>   run "test_security_groups"... skip
> tests/creation.tftest.hcl... tearing down
> tests/creation.tftest.hcl... fail
> 
> Failure! 2 passed, 1 failed, 2 skipped.
> 
> Error: Unable to find user data in list
> 
>   with data.cloudfoundry_user.users["paul.hirsch@gsa.gov"],
>   on main.tf line 19, in data "cloudfoundry_user" "users":
>   19: data "cloudfoundry_user" "users" {
> 
> Given name paul.hirsch@gsa.gov not in the list of users

mogul:
[Many tokens **died** to bring us this information.](https://getyarn.io/yarn-clip/431466c6-2001-4689-87ec-75e9c3f0bf44/gif#p783kV98.copy)

image

rahearn:
@mogul I believe things should work again if you replace every instance of `paul.hirsch` with `anna.mercer`

rahearn (approved):
good to go once the test is passing

rahearn:
> @mogul I believe things should work again if you replace every instance of `paul.hirsch` with `anna.mercer`

better yet, this should be it's own PR: https://github.com/GSA-TTS/terraform-cloudgov/pull/101

PR #103, opened 2026-02-18
Make resource dependencies in client code more explicit and simpler to read

mogul:
## 🛠 Summary of changes

Reduces the need to explicitly pass .id and .name attributes to modules. Just pass modules a resource, and let duck-typing extract the resource attribute values the module needs! This change eliminates redundant data resource lookups (aka cloud.gov API calls), which makes plans and applies faster. (For the permitting project, there were a ton of these implied via the spiffworkflow module; this likely shaves off 10 seconds or so.) Plus it makes reading and reasoning about code easier by keeping resource dependencies explicit rather than sending needed attributes on a side-trip through string vars.

* Adds a new variable "space" to various modules so that you can just pass a `cloudfoundry_space` resource (or data resource!) directly
  * This makes use of Terraform's support for duck-typing
  * This is backward compatible with the community Terraform provider resources, for anyone still using them (I think FAC does in a few places)
* Adds tests for the new methods for specifying the space
* Marks the old parameters as deprecated in plan output to lay ground for removing them in the future
  * I'm intentionally breaking backward compatibility on the spiffworkflow module because there're no users other than the permitting project, and we can pull this same change from `main` into the branch we're using and update the client at the same time, post-merge.

## 📜 Testing Plan

- Run the tests

cc @Mwindo FYI

### DIFF
--- a/app_route/main.tf
+++ b/app_route/main.tf
@@ -1,9 +1,18 @@
+check "deprecated_space_vars" {
+  assert {
+    condition     = var.cf_org_name == "" && var.cf_space_name == ""
+    error_message = "The cf_org_name and cf_space_name variables are deprecated. Use the `space` variable instead to pass a cloudfoundry_space resource directly."
+  }
+}
+
 data "cloudfoundry_org" "org" {
-  name = var.cf_org_name
+  count = var.space == null ? 1 : 0
+  name  = var.cf_org_name
 }
 data "cloudfoundry_space" "space" {
-  name = var.cf_space_name
-  org  = data.cloudfoundry_org.org.id
+  count = var.space == null ? 1 : 0
+  name  = var.cf_space_name
+  org   = data.cloudfoundry_org.org[0].id
 }
 
 data "cloudfoundry_domain" "domain" {
@@ -11,13 +20,14 @@ data "cloudfoundry_domain" "domain" {
 }
 
 locals {
+  space_id = var.space != null ? var.space.id : data.cloudfoundry_space.space[0].id
   destinations = (length(var.app_ids) == 0 ? null : [
     for dest in var.app_ids : { app_id = dest }
   ])
 }
 resource "cloudfoundry_route" "app_route" {
   domain = data.cloudfoundry_domain.domain.id
-  space  = data.cloudfoundry_space.space.id
+  space  = local.space_id
   host   = var.hostname
   path   = var.path
 
--- a/application/main.tf
+++ b/application/main.tf
@@ -9,9 +9,20 @@ data "external" "app_zip" {
   }
 }
 
+check "deprecated_cf_space_name" {
+  assert {
+    condition     = var.cf_space_name == ""
+    error_message = "The cf_space_name variable is deprecated. Use the `space` variable instead to pass a cloudfoundry_space resource directly."
+  }
+}
+
+locals {
+  space_name = var.space != null ? var.space.name : var.cf_space_name
+}
+
 resource "cloudfoundry_app" "application" {
   name       = var.name
-  space_name = var.cf_space_name
+  space_name = local.space_name
   org_name   = var.cf_org_name
 
   path             = "${path.module}/${data.external.app_zip.result.path}"
@@ -37,9 +48,8 @@ resource "cloudfoundry_app" "application" {
 module "route" {
   source = "../app_route"
 
-  cf_org_name   = var.cf_org_name
-  cf_space_name = var.cf_space_name
-  domain        = var.domain
-  hostname      = coalesce(var.hostname, var.name)
-  app_ids       = [cloudfoundry_app.application.id]
+  space    = var.space
+  domain   = var.domain
+  hostname = coalesce(var.hostname, var.name)
+  app_ids  = [cloudfoundry_app.application.id]
 }
--- a/application/main.tf
+++ b/application/main.tf
@@ -16,14 +16,35 @@ check "deprecated_cf_space_name" {
   }
 }
 
+data "cloudfoundry_org" "org" {
+  count = var.space == null ? 1 : 0
+  name  = var.cf_org_name
+}
+
+data "cloudfoundry_space" "space" {
+  count = var.space == null ? 1 : 0
+  name  = var.cf_space_name
+  org   = data.cloudfoundry_org.org[0].id
+}
+
 locals {
-  space_name = var.space != null ? var.space.name : var.cf_space_name
+  space = var.space != null ? var.space : {
+    id   = data.cloudfoundry_space.space[0].id
+    name = var.cf_space_name
+  }
 }
 
 resource "cloudfoundry_app" "application" {
   name       = var.name
-  space_name = local.space_name
+  space_name = local.space.name
   org_name   = var.cf_org_name
+
+  lifecycle {
+    precondition {
+      condition     = var.space != null || var.cf_space_name != ""
+      error_message = "You must provide either the `space` variable or the deprecated `cf_space_name` variable."
+    }
+  }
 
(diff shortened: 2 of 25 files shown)


app_route/main.tf
```diff
@@ -1,23 +1,33 @@
+check "deprecated_space_vars" {
```
rahearn, line 1: [praise] love this use of `check` I'll have to remember it the next time I deprecate some variables

app_route/main.tf
```diff
@@ -1,23 +1,33 @@
...
+  name  = var.cf_space_name
+  org   = data.cloudfoundry_org.org[0].id
 }
 
 data "cloudfoundry_domain" "domain" {
   name = var.domain
 }
 
 locals {
+  space_id = var.space != null ? var.space.id : data.cloudfoundry_space.space[0].id
```
rahearn, line 23: [suggestion (non-blocking)] would a try be more clear than the ternary here?


```suggestion
  space_id = try(var.space.id, data.cloudfoundry_space.space[0].id)
```

rahearn (commented):
Love this for `app_route` - I'm not necessarily opposed for the modules that operated on `cf_space_id` already, but I'm not sure I see the benefit. Is it just for consistency between modules?

Also, my comment on `try` vs `ternary` is appropriate for all of the modules, so if you change it in one place please update everywhere

mogul:
It's kind of hard to explain, but: Sometime between when we were writing a lot of HCL for FAC and earlier systems, and now, we've had to start stitching together `.id` and `.name` everywhere that resources and modules refer to each other explicitly. 

An analogy: It feels like instead of using an object model with an ORM, where the ORM handles persistence of attributes and relationships in the database layer semi-invisibly, we are explicitly managing fkeys in the app layer. So instead of cleanly expressing A(B), we are having to say A(B.id) or A(B.name) everywhere. And instead of cleanly expressing A -> B, we are having to express A -B.id-> B.

app_route/main.tf
mogul: Terraform will throw an invalid reference error when it tries to evaluate the value for the second arg to `try`. Ternaries explicitly do lazy evaluation on the fallback branch, so references are only evaluated when that branch is followed.

mogul:
> It's kind of hard to explain, but: Sometime between when we were writing a lot of HCL for FAC and earlier systems, and now, we've had to start stitching together `.id` and `.name` everywhere that resources and modules refer to each other explicitly.


BTW I think this is also us downstream mimicking a pattern where the resources in the official provider seem to be a lot more picky about what values they accept for attributes, vs the more permissive behavior in the community provider. It's just like what I'm correcting here, where they demand a string. I haven't known enough about Terraform provider mechanics to take on figuring out if that's really the case to suggest how it could be better, but maybe with Claude I could do it. 🤔

mogul:
> BTW I think this is also us downstream mimicking a pattern where the resources in the official provider seem to be a lot more picky about what values they accept for attributes, vs the more permissive behavior in the community provider. It's just like what I'm correcting here, where they demand a string. I haven't known enough about Terraform provider mechanics to take on figuring out if that's really the case to suggest how it could be better, but maybe with Claude I could do it. 🤔

In other words, I'm still 😠 at this change, which seemed totally unnecessary. If we can specify the space by GUID instead of name, there's no need for the app resource definition to be aware of the org at all! Once they made this change we ended up passing the org and space name around a lot more. 

image

mogul:
On the failing tests: `check` blocks cause warnings during `plan` and `apply`, but are errors in the test context. 😩 Since other modules depend on these changes those errors are propagating to other modules. I'm investigating whether there's a way around that.

rahearn:
> In other words, I'm still 😠 at this change

yeah - I do think that some of that is driven by `v2` -> `v3` api changes, rather than the terraform directly, but I agree with your point as to the result as it is exposed to end users.

It looks like the `check` blocks are causing test failures, so those tests likely need to be rewritten (can probably update them to be `command = plan` and check for the failure) and test the "good" path just via the space object.

Other than that, I'm 👍🏻 to this change

mogul:
Finally got back around to taking care of that, @rahearn; take another look?

mogul:
A nice thing I learned in the course of fixing these tests: If you mock at the provider level, it auto-mocks resources which that provider would manage!

mogul:
> A nice thing I learned in the course of fixing these tests: If you mock at the provider level, it auto-mocks resources which that provider would manage!

Turns out this was just another clanker lie. Resource-level mocks restored, tests passing. 

I didn't actually break any clients using these modules, because I kept the deprecated vars and just emitted the check-warning. However because I changed some internals where the modules use each other, and those module references use a relative path... I'm unclear on the behavior if, say, I request `drupal` module `<old>`; what version of module `database` will it use internally? 🤔

mogul:
(If you think that's fine, this is ready to merge. Otherwise we need to either address it, or do a major version bump.)

mogul:
Opus alleges:

image

rahearn:
> Opus alleges:

opus is correct here - gsa-tts/terraform-cloudgov is the actual module, and the `ref` controls the entire module. `//drupal` is a submodule, and can reference other submodules in the same ref via relative paths

application/variables.tf
```diff
@@ -5,7 +5,14 @@ variable "cf_org_name" {
...
   type        = string
-  description = "cloud.gov space name"
+  description = "cloud.gov space name (deprecated; use `space` instead.)"
+  default     = ""
+}
+
+variable "space" {
+  type        = object({ id = string, name = string })
+  description = "A cloudfoundry_space resource (or any object with .id and .name). When provided, cf_space_name is not used."
+  default     = null
```
rahearn, line 15: [suggestion (nonblocking)] should we add a validation here to ensure we're getting either a space or a space_name? For example:

```
validation {
  condition = var.space != null || var.cf_space_name != ""
  error_message = "You must provide a space"
}
```

clamav/main.tf
```diff
@@ -25,9 +36,8 @@ resource "cloudfoundry_app" "clamav_api" {
 module "route" {
   source = "../app_route"
 
-  cf_org_name   = var.cf_org_name
-  cf_space_name = var.cf_space_name
-  domain        = "apps.internal"
-  hostname      = var.name
-  app_ids       = [cloudfoundry_app.clamav_api.id]
+  space    = var.space
```
rahearn: [issue] this won't work if the end-user is still using the deprecated `cf_space_name`

rahearn (changes requested):
I'm not sure what the best way to deal with that clamav setup is. Maybe a conditional data source to get the space if the user passed in cf_space_name that is skipped if they passed in `var.space`?

application/main.tf
```diff
@@ -16,15 +16,36 @@ check "deprecated_cf_space_name" {
...
 
 resource "cloudfoundry_app" "application" {
   name       = var.name
-  space_name = local.space_name
+  space_name = local.space.name
   org_name   = var.cf_org_name
 
+  lifecycle {
+    precondition {
+      condition     = var.space != null || var.cf_space_name != ""
```
rahearn, line 44: [question] any reason you did this as a precondition instead of a variable validation?

rahearn (approved):
🚢

application/main.tf
mogul: 🤖🔮:
image

rahearn: This is no longer true as of terraform 1.9.

rahearn: https://www.hashicorp.com/en/blog/terraform-1-9-enhances-input-variable-validations

mogul: It's only a robot, Ryan.

PR #105, opened 2026-02-23
Add PROXYROUTE to Spiff frontend

Mwindo:
After this is merged, point https://github.com/GSA-TTS/pic-blm-cxworks/pull/692 to use the resulting `custom-spiffworkflow-connector-stable` commit hash and test that PR in dev.


btylerburton (approved):

Mwindo:
I am going to optimistically merge this and update the image ref [pic-blm-side](https://github.com/GSA-TTS/pic-blm-cxworks/pull/692)--if everything works, we are good to go! If not, we can iterate and even revert `custom-spiffworkflow-connector-stable` if need be.

PR #107, opened 2026-03-17
SpiffWorkflow: Make .zip creation the caller's responsibility

mogul:
## 🛠 Summary of changes

Deploys often fail with a missing .zip file because I tried to cram a build step into what should just be a deploy step. This PR attempts to address this historic injustice.

It also updates and adds a bunch of docs, and tries to make everything more robust via validations and expanded tests.

## 📜 Testing Plan

- Run the tests

### DIFF
--- a/spiffworkflow/build-for-cloudfoundry.sh
+++ b/spiffworkflow/build-for-cloudfoundry.sh
@@ -196,21 +196,28 @@ else
   echo "Downloading source code (trying tag, branch, generic) for ref: ${GIT_REF}..."
   TMP_ZIP="${DOWNLOAD_ZIP}.tmp"
   rm -f "$TMP_ZIP"
-  set +e
-  curl -fsSL -o "$TMP_ZIP" "${GIT_URL}/archive/refs/tags/${GIT_REF}.zip" || \
-  curl -fsSL -o "$TMP_ZIP" "${GIT_URL}/archive/refs/heads/${GIT_REF}.zip" || \
-  curl -fsSL -o "$TMP_ZIP" "${GIT_URL}/archive/${GIT_REF}.zip"
-  CURL_RC=$?
-  set -e
-  if [ $CURL_RC -ne 0 ]; then
+
+  DOWNLOAD_URLS=(
+    "${GIT_URL}/archive/refs/tags/${GIT_REF}.zip"
+    "${GIT_URL}/archive/refs/heads/${GIT_REF}.zip"
+    "${GIT_URL}/archive/${GIT_REF}.zip"
+  )
+
+  DOWNLOADED=false
+  for url in "${DOWNLOAD_URLS[@]}"; do
+    if curl -fsSL -o "$TMP_ZIP" "$url" 2>/dev/null; then
+      DOWNLOADED=true
+      break
+    fi
+  done
+
+  if [ "$DOWNLOADED" = false ]; then
     echo "ERROR: Could not download a valid archive for ref '${GIT_REF}' from ${GIT_URL}" >&2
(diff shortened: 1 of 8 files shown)


spiffworkflow/build-for-cloudfoundry.sh
```diff
@@ -0,0 +1,383 @@
...
+fatal() { FAILURE_REASON="$1"; echo "ERROR: $1" >&2; exit 1; }
+
+# This script prepares a deployment zip for the SpiffWorkflow backend application
+# for buildpack-based deployment on Cloud Foundry (cloud.gov).
+# 
+# It handles:
+# - Downloading the source code from GitHub at the specified reference
+# - Copying process models from a local directory
+# - Generating requirements.txt using uv
+# - Optionally including supplemental scripts (init process, profile hooks)
```
Mwindo: I think this comment is outdated (it does not include init process and profile hooks)

spiffworkflow/build-for-cloudfoundry.sh
```diff
@@ -0,0 +1,383 @@
...
+
+# Required parameters
+if [ $# -lt 3 ] || [ $# -gt 5 ]; then
+  fatal "Usage: $0 <output_zip> <backend_gitref> <process_models_path> [python_version] [scripts_path]"
+fi
+
+# Parse arguments
+PACKAGE_PATH="$1"
+# Second argument may be a composite like: github.com/sartography/spiff-arena?ref=v1.1.2
+# or it may still be a simple ref (legacy usage). We derive GIT_URL and GIT_REF.
```
Mwindo: This comment is outdated: we just do the composite now.

spiffworkflow/variables.tf
```diff
@@ -75,10 +75,15 @@ variable "backend_deployment_method" {
   }
 }
 
-variable "backend_gitref" {
-  description = "Git reference (branch, tag, or commit hash) for the upstream spiffworkflow-backend source. Only used when backend_deployment_method = 'buildpack'."
+variable "backend_zip_path" {
```
Mwindo, line 78: Just a note for later that we will probably want to move to s3 eventually. (Probably better than using a GitHub artifact given that we might switch to GitLab at some point.)

Mwindo (approved):
This seems reasonable--the proof will be in the pudding! Thanks @mogul.

spiffworkflow/variables.tf
mogul: [GitLab CI has its own equivalent.](https://docs.gitlab.com/ci/jobs/job_artifacts/) (I expect the main work of migrating will be in our workflows in any case.)

spiffworkflow/build-for-cloudfoundry.sh
mogul:
```suggestion
# Second argument should be a composite like: github.com/sartography/spiff-arena?ref=v1.1.2
# We derive GIT_URL and GIT_REF.
```

spiffworkflow/build-for-cloudfoundry.sh
mogul: We do support including additional scripts, but we don't want to steer people toward init process and profile hooks in particular.

```suggestion
# - Optionally including supplemental scripts
```

mogul:
[I successfully deployed this in the `dev` environment with a `workflow_dispatch`.](https://github.com/GSA-TTS/pic-blm-cxworks/actions/runs/23270711985/job/67662584128) I'm going to merge this here, since it already got an approved review, and then merge https://github.com/GSA-TTS/pic-blm-cxworks/pull/741

PR #108, opened 2026-03-19
fix: make SpiffWorkflow build zip reproducible for stable Terraform hashes

mogul:
- Fetch commit timestamp from GitHub API for SOURCE_DATE_EPOCH-style
  reproducibility; fall back to fixed epoch if unavailable
- Add --no-header to uv pip compile to strip timestamp comment
- Normalize all file mtimes to the commit date before zipping
- Use zip -X to strip extended attributes and TZ=UTC for consistency

This ensures the same inputs always produce the same zip hash, so
Terraform plan/apply correctly detects (or skips) real content changes
based on zip hash.


btylerburton (commented):
Very clever solve with a solid fallback!

PR #109, opened 2026-03-24
Update Trivy actions past compromised versions

mogul:
Looks like [we only tried to use Trivy once since the compromise on March 19th](https://github.com/GSA-TTS/terraform-cloudgov/actions/workflows/pull-and-publish-images.yml), and [that invocation failed because the action had already been pulled](https://github.com/GSA-TTS/terraform-cloudgov/actions/runs/23396524562/job/68060675768).

### DIFF
--- a/.github/workflows/pull-and-publish-images.yml
+++ b/.github/workflows/pull-and-publish-images.yml
@@ -39,7 +39,7 @@ jobs:
         run: docker pull ${{ matrix.image.name }}
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.29.0
+        uses: aquasecurity/trivy-action@0.35.0
         env:
           TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db,ghcr.io/aquasecurity/trivy-db
           TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db,ghcr.io/aquasecurity/trivy-java-db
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-and-publish-images.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         run: docker pull ${{ matrix.image.name }}
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.29.0
+        uses: aquasecurity/trivy-action@0.35.0
```
rahearn: [suggestion] can we please use the sha version for this action, rather than the tag - that'll help avoid a repeat

rahearn (changes requested):

.github/workflows/pull-and-publish-images.yml
mogul: Done. I'm setting up Renovate for the whole org shortly.

.github/workflows/pull-and-publish-images.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         run: docker pull ${{ matrix.image.name }}
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.29.0
+        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1
```
mogul:
```suggestion
        uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # 0.35.0
```

rahearn (approved):

PR #116, opened 2026-04-06
Add an enable_hsts variable ...

Mwindo:
... so we can make use of this: https://github.com/sartography/spiff-arena/commit/41a7620d8aadb39c9d960478102bf9fdcf25e678

Related to:


Mwindo:
Whoops this was silly.

PR #122, opened 2026-04-20
Land PIC team's SpiffWorkflow module work on main

mogul:
## 🛠 Summary of changes
This lands months of work with the SpiffWorkflow module on `main`. Although it seems like a big change, the reality is that the PIC team is the only downstream consumer. 

Note that I rebased the branch onto main to make this PR. This means that along with our update to the spiff-arena codebase, hash IDs have changed! We should update [the versions in use in pic-blm-cxworks](https://github.com/GSA-TTS/pic-blm-cxworks/blob/14c98f662a95a3cb46d9d42b06ef2ab4d742d0a7/.github/spiffworkflow-versions.env#L7) accordingly as soon as this is merged... maybe using an actual tag this time!


.github/workflows/pull-and-publish-images.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             short-name: spiffarena-backend
           - name: ghcr.io/sartography/spiffworkflow-frontend:v2.2.0
             short-name: spiffarena-frontend
-          - name: ghcr.io/sartography/connector-proxy-demo:v2.2.0
+          - name: ghcr.io/gsa-tts/spiffworkflow-connector:latest
             short-name: spiffarena-connector
```
Mwindo: Non-blocking question: Does anyone use these? When should these be used vs. the images in https://github.com/GSA-TTS/spiff-arena?

Mwindo (dismissed):
I did a quick pass and it seems reasonable. Thanks @mogul!

.github/workflows/pull-and-publish-images.yml
mogul: This was originally created when we didn't have a local fork and were trying to  avoid long-term ownership. Now it's pretty clear we need our local fork to address compliance or security stuff ahead of the upstream, so these should change to use the fork's images. Good catch!

mogul: Because of how we've done merges it looks like tagging events were skipped by the build-images GHA. I explicitly [ran the workflow pointing at v2.2.0 just now](https://github.com/GSA-TTS/spiff-arena/actions/runs/24736611840), and will push an update here that uses the newly tagged images once that completes.

mogul: Oh wait, we don't want v2.2.0, because that's the upstream tag. We want "the TTS-flavored version of the most recent upstream release that we're using at time X". I guess I could start a tag convention in `GSA-TTS/spiff-arena` like `v2.2.0-gsatts-<date>`. What do you think?

mogul: You know what, I'm overthinking this. We already have tags like [this one](https://github.com/GSA-TTS/spiff-arena/pkgs/container/spiffworkflow-frontend/804935444?tag=main-2026-04-17_20-30-12-4c42bdc); we should just use them.

.github/workflows/pull-and-publish-images.yml
```diff
@@ -19,11 +19,11 @@ jobs:
       fail-fast: false
       matrix:
         image:
-          - name: ghcr.io/sartography/spiffworkflow-backend:v2.2.0
+          - name: ghcr.io/GSA-TTS/spiffworkflow-backend:main-2026-04-17_20-30-12-4c42bdc
```
btylerburton, line 22: should these all be pulling from `latest`

btylerburton (dismissed):
Looks good outside that one comment about keeping images up to date.

Mwindo:
Something with our service connector was not working, so we thought, "Maybe this change to the service binding had something to do with it somehow?" It did not. Putting a comment here just to explain why it is in the history.

Screenshot 2026-04-22 at 9 26 43 AM

PR #127, opened 2026-07-16
Push images via scopeo to publish all upstream image formats

btylerburton:
## 🛠 Summary of changes

Refactor republish workflow for 3rd party images to pass all architecture formats along to ghcr

Current workflow was only forwarding the amd64 ones.

Test run here: https://github.com/GSA-TTS/terraform-cloudgov/actions/runs/29535192375


mogul (approved):
Skopeo! :til:

